PR #14, opened 2019-09-04
added utf8 support

Buggaboo:
Changes
======

* Something changed on the SDK, it no longer infers the type of the `fromAddress` and `allocate` methods.
* I changed the test to add notes with utf-8 strings as bytes in decimal representation

### DIFF
--- a/lib/src/box.dart
+++ b/lib/src/box.dart
@@ -125,25 +125,25 @@ class Box<T> {
         }
 
         // transform flatbuffers byte array into memory area for C, with a length of a multiple of four
-        Pointer<Uint8> bufferPtr = allocate(count: ((buffer.length + 3.0) / 4.0).toInt() * 4);
+        Pointer<Uint8> bufferPtr = Pointer<Uint8>.allocate(count: ((buffer.length + 3.0) / 4.0).toInt() * 4);
         for(int i = 0; i < buffer.length; ++i)
             bufferPtr.elementAt(i).store(buffer[i] as int);
 
         // put object into box and free the buffer
-        checkObx(bindings.obx_box_put(_objectboxBox, propVals[_idPropIdx]["value"], fromAddress(bufferPtr.address), buffer.length, putMode));
+        checkObx(bindings.obx_box_put(_objectboxBox, propVals[_idPropIdx]["value"], Pointer<Void>.fromAddress(bufferPtr.address), buffer.length, putMode));
         bufferPtr.free();
     }
 
     getById(int id) {
-        Pointer<Pointer<Void>> dataPtr = allocate();
-        Pointer<Int32> sizePtr = allocate();
+        Pointer<Pointer<Void>> dataPtr = Pointer<Pointer<Void>>.allocate();
+        Pointer<Int32> sizePtr = Pointer<Int32>.allocate();
 
         // get element with specified id from database
         Pointer<Void> txn = bindings.obx_txn_read(_store.ptr);
         check(txn != null && txn.address != 0);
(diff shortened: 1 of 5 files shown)


lib/src/ffi/cstring.dart
```diff
@@ -1,24 +1,36 @@
...
 
-    CString(String dartStr) {                                   // if this constructor is used, ".free" needs to be called on this instance
-        _ptr = allocate(count: dartStr.length + 1);
-        for(int i = 0; i < dartStr.length; ++i)
-            _ptr.elementAt(i).store(dartStr.codeUnitAt(i));
-        _ptr.elementAt(dartStr.length).store(0);
+    // if this constructor is used, ".free" needs to be called on this instance
+    CString(String dartStr) {
+        List<int> ints = encodeUtf8(dartStr);
+        var utf8Str = Uint8List.fromList(ints);
```
greenrobot: I think we could get rid of this call if we cast each int to an uint8 in the for loop that follows? Should be a little more efficient?

lib/src/box.dart
```diff
@@ -125,25 +125,25 @@ class Box<T> {
         }
 
         // transform flatbuffers byte array into memory area for C, with a length of a multiple of four
-        Pointer<Uint8> bufferPtr = allocate(count: ((buffer.length + 3.0) / 4.0).toInt() * 4);
+        Pointer<Uint8> bufferPtr = Pointer<Uint8>.allocate(count: ((buffer.length + 3.0) / 4.0).toInt() * 4);
```
greenrobot: I'm getting compile errors with those:

```
lib/src/common.dart:8:50: Error: Method not found: 'Pointer.allocate'.
        Pointer<Int32> majorPtr = Pointer<Int32>.allocate(), minorPtr = Pointer<Int32>.allocate(), patchPtr = Pointer<Int32>.allocate();
                                                 ^^^^^^^^
```
What's your Dart version?
```
$ dart --version
Dart VM version: 2.4.1
```
Maybe we have to increase the Dart version requirement?

nalenz-objectbox: I can confirm @greenrobot's error. If you look at the Dart FFI source file at `/usr/lib/dart/lib/ffi/ffi.dart` on Linux, you can see that the `allocate` function is not a static member of the `Pointer` class, but just outside of it.

Yet, the [FFI SQLite example](https://github.com/dart-lang/sdk/tree/master/samples/ffi/sqlite) uses the presumably static method `Pointer.allocate` as well, despite the underlying source file clearly being different.

lib/src/ffi/cstring.dart
Buggaboo: Sure

lib/src/box.dart
Buggaboo:
```dart --version
Dart VM version: 2.5.0-dev.2.1 (Thu Aug 15 16:05:39 2019 +0200) on "macos_x64"
````
I drink what [homebrew](https://brew.sh/) offers me.

greenrobot: Uh, so FFI plans to change that!? Is there any way to support both version? If we cannot make this work for both, I think we should stick to the latest release version.

Buggaboo: I reverted the parts that are related to the `fromAddress` and `allocate` methods

nalenz-objectbox:
Great! Unfortunately the folder structure for the next version of this binding (0.2), currently in the `dev` branch, changed quite a lot, so I decided to merge your changes manually and close this PR. Also, because regarding FFI, Dart 2.4.1 and 2.5.0 are completely incompatible, we'll have a second branch called `dev-2.5`. Only it will contain a Flutter demo with `Pointer.allocate` instead of `allocate` etc.

PR #16, opened 2019-09-10
0.2 Release

greenrobot:


CHANGELOG.md
```diff
@@ -1,4 +1,14 @@
 0.1.0 (2019-09-03)
 ------------------
 * Minimal Store setup
-* Minimal Box with put and get
\ No newline at end of file
+* Minimal Box with put and get
+
+0.2.0 (2019-09-10)
```
greenrobot: newest version should go to the top

CHANGELOG.md
```diff
@@ -1,4 +1,14 @@
...
+------------------
+* UTF-8 support for Store and Box
```
greenrobot: please say thanks to the contributor and link the PR

examples/flutter/objectbox_demo_desktop/lib/main.dart
```diff
@@ -0,0 +1,106 @@
...
+                _status += "cannot get, as nothing was put in this session yet\n";
+                return;
+            }
+            _status += "fetched note: ${_box.get(_lastPutId)}\n";
+        });
+    }
+
+    void _testPut() {
+        setState(() {
+            _lastPutId = _box.put(Note.construct("Hello"));
```
greenrobot-team: I know little about Flutter, but I was in 1 talk at droidcon about 60FPS in Flutter 😀: and the guy said inside `setState()` no computation should be done, only mutation (ok, [the docs](https://api.flutter.dev/flutter/widgets/State/setState.html) say this, too). So put the note outside of `setState()` and only assign the given ID inside?

Probably applies to the code in `_testGet()` as well.

README.md
```diff
@@ -36,10 +36,10 @@ Getting started
 ---------------
 To try out the demo code in this repository, follow these steps:
 
-1. Install [objectbox-c](https://github.com/objectbox/objectbox-c) system-wide: `bash <(curl -s https://raw.githubusercontent.com/objectbox/objectbox-c/master/download.sh)` (answer Y when it asks about installing to /usr/lib).
-2. Back in this repository, run `pub get` to download all Dart dependencies.
-3. Finally run `dart test/test.dart` to start the demo script.
-   Note that, as fairly recent language features are used, the minimal required Dart version is 2.2.2.
+1. Install [objectbox-c](https://github.com/objectbox/objectbox-c) system-wide: `curl -s https://raw.githubusercontent.com/objectbox/objectbox-c/master/download.sh > /tmp/download.sh && bash /tmp/download.sh 0.7 && rm /tmp/download.sh` (answer Y when it asks about installing to /usr/lib).
+2. Back in this repository, run `pub get` in each of the following directories: `objectbox`, `objectbox_model_generator`, `objectbox_test`. If you just want to try out the tests, running it in `objectbox_test` is enough.
```
vaind: The new branch doesn't follow conventions https://dart.dev/tools/pub/package-layout or rather it creates three packages.

What is the reason it must be split in such a way instead having a single package?
* lib instead of objectbox
* bin/binding-generator instead of objectbox_model_generator
* test instead of objectbox_test

README.md
```diff
@@ -59,37 +59,37 @@ New (not yet persisted) objects typically have _Id_ value of `0` or `null`: call
...
+As specified in step 3 of the _Getting started_ section, Dart's _build\_runner_ and _source\_gen_ are currently used and the generator will be extended to automatically manage the meta model IDs in the future.
 
 ```dart
 import "../lib/objectbox.dart";
+part "test.g.dart";
 
 @Entity(id: 1, uid: 1)
 class Note {
-    @Id(id: 1, uid: 1001, type: Type.Long)
+    @Id(id: 1, uid: 1001)       // automatically always 'int' in Dart code and 'Long' in ObjectBox
```
vaind: IDs should be uint64 (in objectbox), preferrable similara in the sample code as well

greenrobot-team:
Dumb question: the Flutter wiki (https://github.com/flutter/flutter/wiki/Desktop-shells) makes it sound like desktop support is, well, a dumpster fire. Why was the mobile app example replaced with a desktop example?

objectbox_model_generator/lib/src/generator.dart
```diff
@@ -0,0 +1,116 @@
...
+
+            final ${element.name}_OBXDefs = {
+                "model": _${element.name}_OBXModel,
+                "builder": _${element.name}_OBXBuilder,
+                "reader": _${element.name}_OBXReader,
+            };
+        """;
+
+        return ret;
+    }
```
greenrobot-team: Maybe break this method up a little? I feel this getting too long.

objectbox/lib/src/box.dart
```diff
@@ -132,15 +132,20 @@ class Box<T> {
         return propVals[idPropName];
     }
 
+    _inReadTransaction(fn) {
+        Pointer<Void> txn = bindings.obx_txn_read(_store.ptr);
+        checkObxPtr(txn, "failed to created transaction");
+        var ret = fn();
+        checkObx(bindings.obx_txn_close(txn));
```
greenrobot-team: Is there something like a finalizer (e.g. finally-block in Java or defer in Swift) that can make sure this executes, even if `fn()` throws?

greenrobot-team: Looks like there is Finally, like in Java. https://dart.dev/guides/language/language-tour#finally

objectbox_test/test/test.dart
```diff
@@ -1,32 +1,53 @@
...
+            final text = "😄你好";
+            final TestEntity inst = box.get(box.put(TestEntity.construct(text)));
+            expect(inst.text, equals(text));
+        });
+
+        // TODO: test box.putMany, box.getMany, box.getAll
+    });
+
+    tearDown(() {
+        store.close();
```
greenrobot-team: If `setUp()` fails and `store` is `null` would this crash?

greenrobot-team:
Went through all commits (skipped dropped mobile example and the C library stuff I'm not familiar with), added comments. Great!

Looks like Ivan had a look as well. As I'm not around the rest of the week, I guess all follow-up questions go to him. 💃

README.md
nalenz-objectbox: For this, I generally stuck to how Dart's [json_serializable](https://github.com/dart-lang/json_serializable) organizes its files. It uses source_gen as well (just like objectbox_model_generator), so it seemed like quite a good reference for some first orientation. Also, it subdivides the repository into three subprojects as well, each with its own `pubspec.yaml`: `example` (our `objectbox_test`), `json_annotation` (our `objectbox`) and `json_serializable` (our `objectbox_model_generator`). 

Unfortunately, source_gen and build_runner are rather poorly documented for the use case of generating source code for annotations, so it's hard to find good resources on how to better organize the directory structure. Apart from that, I also think that the model generator needs to have its own Dart package in order for [`build.yaml`](https://github.com/objectbox/objectbox-dart/blob/dev-2.5/objectbox_model_generator/build.yaml) to work properly.

README.md
nalenz-objectbox: Do you mean I should change the comment to "...and 'uint64' in ObjectBox"?

objectbox_test/test/test.dart
nalenz-objectbox: Yes, it would. Fixed it.

README.md
vaind: Generated code should pass flags: unsigned & ID

Originally I also meant the entity class should use unsigned integer type but I've found out there's none in dart.

greenrobot: FYI: ["int" in Dart](https://api.dartlang.org/stable/2.5.0/dart-core/int-class.html) is the closest type to ObjectBox IDs. The JS limitations are funny though, btw...

greenrobot: > Generated code should pass flags: unsigned & ID

PropertyFlags_ID (1) is enough; haven't seen PropertyFlags_UNSIGNED (8192) with IDs, better not try this?

README.md
vaind: Core packages are not always a best source of organization since they use to have long history, before most conventions even emerged. 

The page I've linked (https://dart.dev/tools/pub/package-layout) is an official recommendation how a package should be structured and as such it would be strongly preferable to follow there are serious reasons we need to diverge. If you say building the generator is (I don't know enough of dart to help you there at the moment), then let's keep the generator as a separate project for now but at least leave the rest as a single one. What do you think?

objectbox_model_generator/lib/src/generator.dart
nalenz-objectbox: I would prefer to keep it like this for now; as the file is more or less linearly constructed, reading the code from top to bottom makes enough sense here. Of course this will need to change in upcoming versions of the binding, but currently, I'd say that splitting is not really necessary yet.

examples/flutter/objectbox_demo_desktop/lib/main.dart
nalenz-objectbox: Yep, makes sense; fixed it.

objectbox/lib/src/box.dart
nalenz-objectbox: Nice, fixed it.

README.md
nalenz-objectbox: Great, I definitely agree; I'll also keep the generator as a separate package for now. Please check out my last commit, I reorganized the file structure similar to how you suggested.

README.md
nalenz-objectbox: Right, I haven't really seen `PropertyFlags_UNSIGNED` before either... Can this conversation be closed then?

Issue #17, opened 2019-09-11
Read and write chunks of memory for CString


nalenz-objectbox:
Currently, [writing strings to](https://github.com/objectbox/objectbox-dart/blob/master/lib/src/ffi/cstring.dart#L13) and [reading strings from](https://github.com/objectbox/objectbox-dart/blob/master/lib/src/ffi/cstring.dart#L26) memory is done character by character. Try to find a more efficient method to do that.

Is mostly related to CString (#7), but also memory chunks in general.

vaind:
[Dart FFI homepage](https://dart.dev/guides/libraries/c-interop) links the following example:

Even though it still iterates, it looks less cumbersome than the current solution so maybe we could switch to that if there's no more efficient solution (consider opening an issue on dart github).

Buggaboo:
I'm on it

vaind:
Thanks @Buggaboo for tackling this :1st_place_medal:

PR #21, opened 2019-09-13
Replaced CString with ffi utf util package

Buggaboo:
Fix for issue #17


lib/src/common.dart
```diff
@@ -15,7 +15,7 @@ class Common {
     }
 
     static String versionString() {
-        return CString.fromPtr(bindings.obx_version_string()).val;
+        return Utf8.fromUtf8(bindings.obx_version_string().cast<Utf8>());
```
vaind: What do you think about changing the definition of `obx_version_string()` in `bindings.dart` to return `Pointer<Utf8>` instead of `Pointer<Uint8>`? So we can get rid of the cast. There are a few functions returning/taking `Pointer<Uint8>` so I guess we can change all of them.

lib/src/common.dart
```diff
@@ -24,7 +24,7 @@ class Common {
 
         int last = bindings.obx_last_error_code();
         int last2 = bindings.obx_last_error_secondary();
-        String desc = CString.fromPtr(bindings.obx_last_error_message()).val;
+        String desc = Utf8.fromUtf8(bindings.obx_last_error_message().cast<Utf8>());
```
vaind: same as above, the cast wouldn't be necessary

vaind:
```suggestion
        String desc = Utf8.fromUtf8(bindings.obx_last_error_message());
```

lib/src/common.dart
vaind:
```suggestion
        return Utf8.fromUtf8(bindings.obx_version_string());
```

lib/src/model.dart
```diff
@@ -31,22 +32,16 @@ class Model {
...
             modelDefinitions.forEach((m) {
                 // start entity
-                var entityName = CString(m["entity"]["name"]);
-                try {
-                    checkObx(bindings.obx_model_entity(_objectboxModel, entityName.ptr, m["entity"]["id"], m["entity"]["uid"]));
-                } finally {
-                    entityName.free();
-                }
+                var entityUtf8 = Utf8.toUtf8(m["entity"]["name"]);
+                var entityNamePointer = entityUtf8.cast<Uint8>();
```
vaind: this cast could also go away
```suggestion
```

lib/src/model.dart
```diff
@@ -31,22 +32,16 @@ class Model {
...
+                checkObx(bindings.obx_model_entity(_objectboxModel, entityNamePointer, m["entity"]["id"], m["entity"]["uid"]));
```
vaind: Note that Utf8.toUtf8 returns malloc-ed string so the resulting pointer must be freed after it has been passed to obx_model_entity()
```suggestion
                try {
                    checkObx(bindings.obx_model_entity(_objectboxModel, entityUtf8, m["entity"]["id"], m["entity"]["uid"]));
                } finally {
                    entityUtf8.free();
                }
```

lib/src/model.dart
```diff
@@ -31,22 +32,16 @@ class Model {
...
-                    var propertyName = CString(p["name"]);
-                    try {
-                        checkObx(bindings.obx_model_property(_objectboxModel, propertyName.ptr, p["type"], p["id"], p["uid"]));
-                        checkObx(bindings.obx_model_property_flags(_objectboxModel, p["flags"]));
-                    } finally {
-                        propertyName.free();
-                    }
+                    var propertyUtf8 = Utf8.toUtf8(p["name"]);
+                    var propertyNamePointer = propertyUtf8.cast<Uint8>();
+                    checkObx(bindings.obx_model_property(_objectboxModel, propertyNamePointer, p["type"], p["id"], p["uid"]));
```
vaind: same as above

lib/src/model.dart
vaind: Hmm seeing all those try-finally-free I'm thinking if we could have a helper function like:

```dart
T withCString<T>(String string, T Function(Pointer<Utf8>) fn) {
    var utf8 = Utf8.toUtf8(string);
    try {
        return fn(utf8);
    } finally {
        utf8.free();
    }
}
```

lib/src/store.dart
```diff
@@ -20,12 +23,8 @@ class Store {
...
             checkObx(bindings.obx_opt_model(opt, model.ptr));
             if (directory != null && directory.length != 0) {
-                var cStr = new CString(directory);
-                try {
-                    checkObx(bindings.obx_opt_directory(opt, cStr.ptr));
-                } finally {
-                    cStr.free();
-                }
+                var cStr = Utf8.toUtf8(directory).cast<Uint8>();
+                checkObx(bindings.obx_opt_directory(opt, cStr));
```
vaind: freeing necessary as well

lib/src/common.dart
Buggaboo: I don't see how, sorry. The `#cast<Utf8>` suggests will be lenient in converting the Function's return type from `Pointer<Uint8>` to `Pointer<Utf8>`, but it isnt. Afaik, the `Utf8` type is a struct, and not a scalar like Uint8.

lib/src/common.dart
Buggaboo: idem dito, unless I overlooked something

lib/src/model.dart
Buggaboo: I assumed the ffi:Utf8 lib already freed the malloc.
The Function signatures are pretty specific, maybe if there's a general pattern, we can definitely use that helper function.

lib/src/store.dart
Buggaboo: In the future, when we can leverage an `extension`, we could have a `Utf8.toUintPointer`.
Or, we could add that to the [ffi util](https://github.com/dart-lang/ffi) and PR.

lib/src/model.dart
vaind: I originally hoped the same but there seem to be no destructors in Dart so smart pointers are out of the question, unfortunately...

lib/src/common.dart
vaind: I'll try to have a after the merge if what I had in mind works

vaind:
thanks again, merging :+1:

Issue #19, opened 2019-09-11
Refactoring


vaind:
- [x] move common.lastErrorString to binding/helpers.dart - it's the only place where it's called from
- [x] since generated `"flatbuffers_id": ${(prop['id'] as int) - 1},` is always `property.Id - 1` it can be removed completely (as in other language bindings)
- [x] remove box.close() - that's not how a store should be closed
- [x] split flatbuffers marshalling out of box.dart
- [x] use `dartfmt -l 120` (and add it to contribution guide?)

PR #23, opened 2019-09-16
Refactoring

nalenz-objectbox:
See #19.


DynamicCounter:
It always gives me an error no matter what example I take: objectbox.g.dart "Target of URI hasn't been generated: 'objectbox.g.dart'.
Try running the generator that will generate the file referenced by the"

Issue #18, opened 2019-09-11
Automatically generate IDs and UIDs for entities and properties


nalenz-objectbox:
- when writing an entity class, it should be enough to specify the respective annotation without any parameters, i.e. `@Entity() class MyEntity { ... }` should suffice
- all instance variables of the entity class shall be incorporated into the final model and only those with special properties (e.g. `@Id()`) need to be annotated
- all IDs and UIDs need to be generated by `objectbox_model_generator`, together with [these rules](https://docs.objectbox.io/advanced/data-model-updates) on how to resolve changes and conflicts
- in contrast to the current situation, this means that the existing generated file (specifically the `_MyEntity_OBXModel` object) needs to be parsed beforehand in order to detect changes

nalenz-objectbox:
The branch `18-automatic-id-generation` now contains the following changes:

- a file called `objectbox-models.json` is generated with a structure similar to ObjectBox Java
  - only merging entity properties is supported right now, indexes and relations are still completely missing
  - the overall file structure of the new generator is heavily influenced by the Go source you provided the link to
- unannotated fields in an `@Entity()` class now behave just like those annotated with `@Property()`

vaind:
Uff, it's huge already. Please create a pull request (`18-...` to `dev`) so that I can review properly and restrain yourself from adding more stuff (if you planned to do so) unless necessary for this issue - it's going to be hard to review as it is now. Thanks

Buggaboo:
At the risk of increasing the lines of code here, are there plans to support other types other than `String` and `int`? Currently my query tests are failing because things like `bool` and `double` are not supported.

vaind:
> At the risk of increasing the lines of code here, are there plans to support other types other than `String` and `int`? Currently my [query tests](https://github.com/Buggaboo/objectbox-dart/tree/query) are failing because things like `bool` and `double` are not supported.

It's not really related to the original goal of this ticket and should rather be covered by #11 - so that we can merge this PR ASAP.

Issue #5, opened 2019-09-03
Basic Query (Builder) support


greenrobot:
Goal: minimal solution including a query builder and a reusable query object. For reference, please check the [Java query API](https://docs.objectbox.io/queries).

- [ ] QueryBuilder to allow "equal" criteria for string and int properties (pass property IDs to identify properties if we don't generate a meta model yet)
- [ ] QueryBuilder allows building a Query object
- [ ] Query object methods: find() and findFirst(). find() returns all matching objects, findFirst() just the first one.

Buggaboo:
I started by wrapping the C functions, I'll share a branch soon.

With regard to designing the interface, I'm thinking of passing a closure to the  Store#query method, to produce a QueryBuilder object like swift.

I think another way to do it idiomatically in dart is by declaring this `query` method as a `factory` that takes a closure that takes instances of `Condition` in the body. Also one can leverage the `..` _cascade operator_, so there's no need to declare those types as part of a builder pattern. And boolean operator overloading for any / join (i.e. respectively `||`, `&&`).

At some point in the future, we'll can leverage `extension` (a la Objc: `category`, swift: `extension`), to bolt on the properties to create conditions.

vaind:
Factory method taking `Condition` is what Go does (not closure, directly a list of conditions). That's probably similar to what you had in mind, isn't it? Maybe it would make the code/usage simpler without the closure?

```
box.Query(Person_.age.greaterThan(18), Person_.name.startsWith("John"))
```

vaind:
A few things to keep in mind:
* consider building queries an expensive operation, the idea is they should be built once and reusable (just changing parameter values)
* there are links (over relation fields) - swift and go take slightly different approaches, with Go creating a condition and passing it to the query and swift chaining the link on the query object (therefore requiring an explicit `build()` call which Go doesn't need)

In any case, having a look at the query-builder implementation in go/swift may help you clear some things up. https://github.com/objectbox/objectbox-go/blob/master/objectbox/querybuilder.go

vaind:
* java - https://docs.objectbox.io/queries
* go - https://golang.objectbox.io/queries
* swift - https://swift.objectbox.io/queries

Buggaboo:
We can't use varargs or reuse method names, that's why I initially thought of passing a closure.

vaind:
Oh, no variadic functions in dart? What about taking a list then? Still less cumbersome than a closure

vaind:
[...]

PR #25, opened 2019-09-25
Automatically generate IDs and UIDs for entities and properties

nalenz-objectbox:
See #18.


.gitignore
```diff
@@ -4,10 +4,10 @@
...
 misc/
 .idea/
-**/*.g.dart
 download/
 lib/*.dll
 lib/*.dylib
 lib/*.so
 lib/*.a
 .vscode/
+**/*.g.dart
```
vaind: Not sure about ignoring the generated files since they're not auto-refreshed on during build, but since it already was ignored, let's keep it for now

lib/src/model.dart
```diff
@@ -34,11 +31,11 @@ class Model {
       // transform classes into model descriptions and loop through them
       modelDefinitions.forEach((m) {
         // start entity
-        var entityUtf8 = Utf8.toUtf8(m["entity"]["name"]);
+        var entityUtf8 = Utf8.toUtf8(m["name"]);
```
vaind: Seeing this as well as rest of the code, I have a feeling it would be much cleaner with a "normal" class-based representation of a model. Nothing fancy is necessary, just the fields that are currently present in the json "entities" path

lib/src/store.dart
```diff
@@ -45,7 +44,7 @@ class Store {
   }
 
   getEntityModelDefinitionFromClass(cls) {
-    return _modelDefinitions[cls]["model"];
+    return _modelDefinitions[cls]["model"]();
```
vaind: the dynamic stuff is pretty hard to reason about without insider information. What's this returning?

test/test.dart
```diff
@@ -3,14 +3,16 @@ import "package:test/test.dart";
...
+@Entity()
 class TestEntity {
-  @Id(id: 1, uid: 1001)
+  @Id()
   int id;
 
-  @Property(id: 2, uid: 1002)
   String text;
 
+  @Property(type: OBXPropertyType.Short)
```
vaind: We don't really need custom type sizes at the moment so let's not complicate things with those now (e.g. validation whether the value fits into short). Just map the built-in types for now

lib/src/box.dart
```diff
@@ -17,14 +17,24 @@ class Box<T> {
...
+  var _idPropName;
 
   Box(this._store) {
     _entityDefinition = _store.getEntityModelDefinitionFromClass(T);
     _entityReader = _store.getEntityReaderFromClass<T>();
     _entityBuilder = _store.getEntityBuilderFromClass<T>();
     _fbManager = new OBXFlatbuffersManager<T>(_entityDefinition, _entityReader, _entityBuilder);
 
-    _objectboxBox = bindings.obx_box(_store.ptr, _entityDefinition["entity"]["id"]);
+    for (int i = 0; i < _entityDefinition["properties"].length; ++i) {
```
vaind: not sth. box should be doing at runtime - id Property is defined by the model

vaind: see my comment in model.dart

vaind:
Also don't forget to rebase/merge changes from dev

objectbox-models.json
```diff
@@ -0,0 +1,38 @@
+{
```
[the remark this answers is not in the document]
nalenz-objectbox: Actually, regarding your second point, this is why I decided to put the file into the root directory:

- assuming you have multiple source files in different directories, which might all define models, separate JSON files would make the act of merging much more difficult, because to make sure the merge is consistent, all other JSON files need to be read and checked regarding conflicts beforehand
- it works fine with `flutter run`. I'm actually not sure whether Flutter also compiles the Dart code like you mentioned, but it seems like at least the current working directory Flutter code is executed from is still the project's root directory

Also, I would've definitely liked if the (or one) generated `.g.dart` file(s) could contain all model definitions, but unfortunately, all of the Dart AST parsing capabilities, presumably provided by `dart:analyzer` are even less well-documented than the generator itself, so it's virtually impossible to get this to run.

.gitignore
nalenz-objectbox: Right, I had committed the `.g.dart` files into the repository before, but when I tried to rebuild them, `build_runner´ warned me that this file would be overwritten. This warning only occurred, when the file was version-controlled, so because it is essentially generated from the JSON file anyway, there's no need to have this redundant information in the repository.

objectbox-models.json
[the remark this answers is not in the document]
vaind: Model.json - you're right that there should be only a single file, what I'm saying is it should be in specific to a single package (tests seem to be standalone). 

Generated code - no AST parsing is necessary at all. The code is already generated right? Just generate the code directly with the information from the model.json file (during build step). 
Go does it slightly differently but just for reference: https://github.com/objectbox/objectbox-go/blob/master/examples/tasks/internal/model/task.obx.go#L63

What prevents representing a model using classes instead of a map<string, dynamic>? 
Something like [this generated code from go](https://github.com/objectbox/objectbox-go/blob/master/examples/tasks/internal/model/task.obx.go#L25) would be (probably) also necessary for queries in Dart as well, see discussion here: https://github.com/objectbox/objectbox-dart/issues/5#issuecomment-531496314.

Are we on the same page?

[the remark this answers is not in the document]
vaind: As an example, the generated code could contain something like the following. 

```dart
class _TestEntity {
  PropertyLong id;
  PropertyString text;
  PropertyInt number;
}

var TestEntity_ = new _TestEntity();
TestEntity_.id = PropertyLong(id: 1, uid: 9211738071025439652, type: 6);
TestEntity_.text = PropertyString(id: 2, uid: 8804670454579230281, type: 9);
TestEntity_.number = PropertyInt(id: 3, uid: 1119420894747302346, type: 3);

```

Something like that would be necessary for Queries anyway. See discussion in  https://github.com/objectbox/objectbox-dart/issues/5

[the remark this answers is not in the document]
vaind: I'm just a little sceptical about maintainability of the binding with all these maps everywhere - all IDE help with refactoring, show usage, etc. is impossible because it's just strings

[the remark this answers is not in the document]
vaind: As a start, a `_getOBXModel()` could be replaced to return an instance of `Entity` that could look sth like:
```dart

class Entity {
  final int id;
  final int uid;
  final string name;
  List<Property> properties;
}
```

You get the idea what the `Property` class would look like.

objectbox-models.json
```diff
@@ -0,0 +1,38 @@
...
+          "flags": 0
+        }
+      ]
+    }
+  ],
+  "lastEntityId": "1:6645479796472661392",
+  "retiredEntityUids": [],
+  "retiredPropertyUids": [],
+  "modelVersion": 1,
+  "minimumParserVersion": 1,
```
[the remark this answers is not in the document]
nalenz-objectbox: Done.

objectbox-models.json
```diff
@@ -0,0 +1,38 @@
...
+          "id": "1:9211738071025439652",
+          "name": "id",
+          "type": 6,
+          "flags": 1
+        },
+        {
+          "id": "2:8804670454579230281",
+          "name": "text",
+          "type": 9,
+          "flags": 0
```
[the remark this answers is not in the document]
nalenz-objectbox: Done.

test/test.dart
nalenz-objectbox: Done.

objectbox-models.json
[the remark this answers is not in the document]
nalenz-objectbox: I made the respective changes to use as few maps as possible; sorry for the large commit for that, but although logically everything stayed mostly the same, it required quite a lot of code adjustments.

lib/src/model.dart
nalenz-objectbox: Done.

lib/src/store.dart
nalenz-objectbox: This stems from the generated `TestEntity_OBXDefs` map. Right now, I don't see a way to use something class-based here as well, because at that point, we need as much flexibility as possible (type-wise) and it seems like this is not possible with e.g. inheritance, but only maps, where the respective type conversions happen later (that is, e.g. [here](https://github.com/objectbox/objectbox-dart/blob/18-automatic-id-generation/lib/src/store.dart#L52); notice the type conversion involving the template parameter).

lib/src/box.dart
nalenz-objectbox: Done.

bin/objectbox_model_generator/lib/src/code_chunks.dart
```diff
@@ -0,0 +1,50 @@
...
+          Map<String, dynamic> r = {};
+          ${readEntity.properties.map((p) => "r[\"${p.name}\"] = inst.${p.name};").join()}
+          return r;
+        }
+
+        const ${name}_OBXDefs = {
+          "getModelEntity": _${name}_OBXModelGetter,
+          "convertMapToInstance": _${name}_OBXBuilder,
+          "convertInstanceToMap": _${name}_OBXReader,
+        };
```
vaind: Could you also change `*_OBXDefs` to be an instance of a class with those three properties? It would have multiple benefits:
* old generated code wouldn't compile if the class has changed
* renames of the properties would be easy - no need to access by a string

lib/src/store.dart
vaind: Implemented in d0cce264fce1babf09e2e3bd158de5e9a42b7788

vaind:
The remaining things before this PR can be merged:
* **rebase/merge the dev branch**
* tests - e.g. generate code for multiple entities, not just one. Swift & Go bindings do tests by comparing the generated code & model to an "expected" version for each `test case` (a folder here: https://github.com/objectbox/objectbox-go/tree/master/test/generator/testdata). No need to go to such extent but at least some tests should be there
* at least some basic docs for the new classes

Buggaboo:
I'm currently working on doing Queries, is the boolean / byte / double / long implementation for generating properties included in this branch?

vaind:
Merging this into dev, please add the missing class docs directly on dev branch @nalenz-objectbox

PR #27, opened 2019-10-03
WIP query

Buggaboo:
Last hurdle: fixing the tests, take inspiration from the [golang tests](https://github.com/objectbox/objectbox-go/blob/master/test/query_test.go#L99).


lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+  // TODO let the programmer decide on the resolution via @Property
+  QueryCondition equal(bool p) {
+    final c  = Condition<int>(ConditionOp._eq, type, (p ? 1 : 0));
+    return QueryCondition(entityId, propertyId, c);
+  }
+
+  QueryCondition operator == (bool p) => equal(p);
+}
+
+class OBXOrderFlag {
```
vaind: Let's skip ordering for now, i.e. remove/comment out/unpublish by prefixing with an underscore. We should handle this in a separate issue.

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+  static final UNSIGNED = 4;
+
+  /// null values will be put last.
+  /// If not specified, by default null values will be put first.
+  static final NULLS_LAST = 8;
+
+  /// null values should be treated equal to zero (scalars only).
+  static final NULLS_ZERO = 16;
+}
+
```
vaind: everything beyond this line seems to be concerned with building the queries so I'd suggest splitting out to `query_builder.dart`

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+enum ConditionOp {
```
vaind: Maybe unpublish (prefix with an underscore) and follow "effective-dart" naming rules for values

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+  Condition(this._op, this._type, this._value, [this._value2 = null]);
+  Condition.fromList(this._op, this._type, this._list);
+
+  int _nullness(Pointer<Void> qbPtr, QueryCondition qc, obx_qb_cond_operator_0_dart_t func) {
+    return func(qbPtr, qc._propertyId);
+  }
+}
+
+class StringCondition extends Condition<String> {
+  List<int> orderFlags;
```
vaind: this should also be removed for the time being, together with the "descending" constructor argument

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+    return new QueryCondition(entityId, propertyId, c);
+  }
+
+  QueryCondition notNull() {
+    final c = Condition<int>(ConditionOp._not_null, null, 0);
+    return new QueryCondition(entityId, propertyId, c);
+  }
+}
+
+class QueryStringProperty extends QueryProperty {
```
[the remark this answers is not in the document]
vaind: Actually, we could just do similar what Go does: `BaseProperty/PropertyBase` and `PropertyString`, etc.
No need to have this "namespaced" for queries only (even though queries are the only thing currently using it)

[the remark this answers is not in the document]
vaind: could even be in its own `property.dart` file

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+  int _nullness(Pointer<Void> qbPtr, QueryCondition qc, obx_qb_cond_operator_0_dart_t func) {
+    return func(qbPtr, qc._propertyId);
+  }
+}
+
+class StringCondition extends Condition<String> {
+  List<int> orderFlags;
+  bool _caseSensitive = false, _withEqual = false;
+
+  StringCondition(ConditionOp op, ConditionType type, String value, String value2, int caseSensitive, int descending)
```
vaind: caseSensitive when used with string conditions is not a `OBXOrderFlag` but a boolean. 

FYI: case-sensitive order flag, i.e. not part of this PR because order isn't, handles only ordering, not conditions (as those are distinct)

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+    final ptr = Pointer<Uint64>.allocate(count: 1);
+    try {
+      checkObx(bindings.obx_query_count(_query, ptr));
+      return ptr.load();
+    }finally {
+      ptr.free();
+    }
+  }
+
+  // TODO does dart have a dtor/finalizer?j
```
vaind: unfortunately, it doesn't... seems there's no way to have this called automatically so you can safely remove the TODO

lib/src/bindings/structs.dart
```diff
@@ -13,6 +14,19 @@ class IDArray {
     _structPtr.elementAt(1).store(ids.length);
   }
 
+  List<int> _ids; // obx_id === uint64_t
+
+  IDArray.fromAddress(int address) {
```
vaind: Seems like we should change `IDArray` to an `ffi.Struct`: see https://github.com/dart-lang/samples/blob/master/ffi/structs/structs.dart#L52
with the C struct looking like this: https://github.com/dart-lang/samples/blob/master/ffi/structs/c/structs.h#L11

`OBX_id_array` is defined here:
https://github.com/objectbox/objectbox-c/blob/master/include/objectbox.h#L319

vaind: also, memory is not being reed using `obx_id_array_free()`

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+  void close() {
+    checkObx(bindings.obx_query_close(_query));
+  }
+
+  // TODO reimplement the same marshalling as in Box.getMany???
+  T findFirst() {
+    final list = findIds(offset:0, limit:1);
+    return (list == null ? null : _box.get(list.first)) as T;
```
vaind: If you really want to implement this - it should be run in a single read transaction

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+  }
+
+  List<int> findIds({int offset=0, int limit=(1<<32)}) {
```
vaind: While I get the point of using limit as a large number, I think we're using limit=0 for "no limit" in all other bindings so we should probably go for the same here

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+    final structPtr = checkObxPtr(bindings.obx_query_find_ids(_query, offset, limit), "find ids");
+    final idArray = IDArray.fromAddress(structPtr.address);
+    return idArray.ids.length == 0 ? null : idArray.ids;
+  }
+
+  List<T> find({int offset=0, int limit=(1<<32)}) {
+    final list = findIds(offset:offset, limit:limit);
+    return list == null ? null : _box.getMany(list);
```
vaind: this way it's a little inefficient so if anything, it should be done in a read transaction & be marked with a big TODO to use `obx_query_find()`

lib/src/annotations.dart
```diff
@@ -3,9 +3,19 @@ class Entity {
...
+ *
+ * Property allows the mapping to be specific. The defaults are
+ * e.g. Int -> Int64, double -> Float64, bool -> Bool.
+ *
+ * Use OBXPropertyType and OBXPropertyFlag values, resp. for type and flag.
+ */
 class Property {
-  final int uid;
-  const Property({this.uid = null});
+  final int uid, type, flag;
```
vaind: Do we really need flag annotations?

vaind: Also I'm not sure I understand why type is necessary for this PR. Maybe it was before the code generator changes ("id generation")? If it's not necessary/used, I'd suggest reverting those changes for now

README.md
```diff
@@ -98,6 +98,44 @@ print("refetched note: ${box.get(note.id)}");
...
+queryNullText.close(); // We have to manually close queries and query builders.
+```
+
+More complex queries can be constructed using `and/or` operators.
+Also there is basic operator overloading support for `equal`, `greater`, `less`, `and` and `or`,
+respectively `==`, `>`, `<`, `&`, `|`.
+
+```dart
+final anyGroup = <QueryCondition>[text.equals("meh"), text.equals("bleh"), text.contains("Hello")];
+final queryAny = box.queryAny(anyGroup).build();
```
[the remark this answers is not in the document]
Buggaboo: Actually, [you inspired this change](https://github.com/objectbox/objectbox-dart/issues/5#issuecomment-531685949).

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+ * contains the enum representation of
+ * the to-be-constructed query builder.
+ * This design allows nested chains inside
+ * the chain.
+ */
+class QueryCondition {
+  bool _root = true;
+  int _entityId, _propertyId;
+  Condition _condition;
+  List<List<QueryCondition>> _dnf; // all
```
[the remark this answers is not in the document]
Buggaboo: Renamed `dnf` to `anyGroup`

lib/src/bindings/structs.dart
Buggaboo: I agree `Struct` would make this prettier. I already tried hacking this the `ffi.Struct` way, but for some reason I couldn't get it right. (wrong imports? wrong version?)
You could give it a try.

Buggaboo: `obx_id_array_free` done.

lib/src/query.dart
Buggaboo: owww... thanks

lib/src/query.dart
Buggaboo: I totally agree.

lib/src/annotations.dart
Buggaboo: No it's new. The `flag` is there to open up support for all the flags like uniqueness constraints etc. My reasoning: Instead of creating a new annotation for each property, one annotation should cover them all.

In the case of the `type`, I can imagine some users who don't need `Int64` for each int. This allows for some specificity.

lib/src/bindings/signatures.dart
```diff
@@ -55,3 +55,70 @@ typedef obx_box_put_native_t = Int32 Function(Pointer<Void> box, Uint64 id, Poin
 typedef obx_box_put_many_native_t = Int32 Function(
     Pointer<Void> box, Pointer<Uint64> objects, Pointer<Uint64> ids, Int32 mode);
 typedef obx_box_remove_native_t = Int32 Function(Pointer<Void> box, Uint64 id);
+
+// no typedef for non-functions
+// typedef obx_err = Int32
+// typedef Pointer<Int8> -> char[]
+// typedef Pointer<Int32> -> int (e.g. obx_qb_cond);
```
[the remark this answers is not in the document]
Buggaboo: Where do I add the link?

lib/src/annotations.dart
Buggaboo: But you're right this, should be outside of this PR.

test/box_test.dart
```diff
@@ -92,12 +99,243 @@ void main() {
...
+      final queryAny0 = box.queryAny(anyGroupCondition0).build();
+      final queryAll0 = box.queryAll(allGroupCondition0).build();
+
+      final q1 = box.query(cond1).build();
+      final q2 = box.query(cond2).build();
+      final q3 = box.query(cond3).build();
+      final q4 = box.query(cond4).build();
+
+      box.query(selfInference1 as QueryCondition);
+      box.query(selfInference2 as QueryCondition);
```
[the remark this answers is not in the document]
Buggaboo: I was testing the syntax / compiler / type inference.

test/box_test.dart
```diff
@@ -92,12 +99,243 @@ void main() {
...
+    test(".count matches of `in`, `contain`, `contains`", () {
+      box.put(TestEntity.initIntegerAndText(1337, "meh"));
+      box.put(TestEntity.initIntegerAndText(1,    "bleh"));
+      box.put(TestEntity.initIntegerAndText(1337, "bleh"));
+      box.put(TestEntity.initIntegerAndText(1337, "blh"));
+
+      final text = TestEntity_.text;
+      final number = TestEntity_.number;
+
+      final qs0 = box.query(text.inList([ "meh" ])).build();
```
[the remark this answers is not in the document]
Buggaboo: I have made the changes.

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+  }
+}
+
+// Construct a tree from the first condition object
+class QueryBuilder<T> {
+  Box<T> _box;
+  Store _store;
+  int _entityId; // aka model id, entity id
+  QueryCondition _queryCondition;
+  Pointer<Void> _queryBuilderPtr;
```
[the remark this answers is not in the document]
Buggaboo: Done

lib/src/box.dart
```diff
@@ -52,9 +53,13 @@ class Box<T> {
 
     // put object into box and free the buffer
     ByteBuffer buffer = _fbManager.marshal(propVals);
-    checkObx(bindings.obx_box_put(
-        _objectboxBox, propVals[_modelEntity.idPropName], buffer.voidPtr, buffer.size, _getOBXPutMode(mode)));
-    buffer.free();
+    try {
```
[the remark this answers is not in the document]
Buggaboo: Will do from now on.

test/box_test.dart
```diff
@@ -92,12 +99,243 @@ void main() {
...
+
+//      final q0 = box.query(d > 0.1).build();
+//      final q1 = box.query(b == false).build();
+//      final q2 = box.query(t > "more").build();
+//      final q3 = box.query(t < "more").build();
+//      final q4 = box.query(d < 0.3).build();
+//      final q5 = box.query(n < 1337).build();
+//      final q6 = box.query(n > 1337).build();
+
+      final q0 = box.query(d.greater(0.1)).build();
```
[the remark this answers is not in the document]
Buggaboo: Done

lib/src/bindings/bindings.dart
```diff
@@ -136,6 +188,70 @@ class _ObjectBoxBindings {
     obx_box_put = objectbox.lookup<NativeFunction<obx_box_put_native_t>>("obx_box_put").asFunction();
     obx_box_put_many = objectbox.lookup<NativeFunction<obx_box_put_many_native_t>>("obx_box_put_many").asFunction();
     obx_box_remove = objectbox.lookup<NativeFunction<obx_box_remove_native_t>>("obx_box_remove").asFunction();
+
+    // query builder
+    obx_qb_create = objectbox.lookup<NativeFunction<obx_query_builder_native_t>>("obx_query_builder").asFunction();
```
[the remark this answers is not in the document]
Buggaboo: Done.

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+  QueryCondition startsWith(String p, {bool descending = false}) {
+    return _op(p, ConditionOp._string_starts, false, descending);
+  }
+
+  QueryCondition contains(String p, {bool caseSensitive = false}) {
+    return _op(p, ConditionOp._string_contains, caseSensitive, false);
+  }
+
+  // TODO rename to something better
+  QueryCondition contain(String p, {bool caseSensitive = false}) {
```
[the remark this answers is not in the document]
Buggaboo: I was wondering what it does.

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+//  QueryCondition operator != (String p) => notEqual(p); // not overloadable
+}
+
+class QueryIntegerProperty extends QueryProperty {
+  QueryIntegerProperty(int entityId, int propertyId) : super(entityId, propertyId);
+
+  static const ConditionType type = ConditionType._int64;
+
+  QueryCondition _op(int p, ConditionOp cop) {
+    final c = IntegerCondition(ConditionOp._eq, type, p, 0);
```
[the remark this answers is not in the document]
Buggaboo: Nice catch, thanks!

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+  QueryCondition notNull() {
+    final c = Condition<int>(ConditionOp._not_null, null, 0);
+    return new QueryCondition(entityId, propertyId, c);
+  }
+}
+
+class QueryStringProperty extends QueryProperty {
+  QueryStringProperty(int entityId, int propertyId) : super(entityId, propertyId);
+
+  static const ConditionType type = ConditionType._string;
```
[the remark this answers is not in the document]
Buggaboo: This was a temporary hack to provide sane type mappings. Since adding those params to `Property` I could take user-defined mappings instead.

Actually, to think of it, `OBXPropertyType` could cover these cases.

lib/src/query.dart
Buggaboo: Oops, I'll separate those.

lib/src/query.dart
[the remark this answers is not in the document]
Buggaboo: Sure, I'll unnamespace them.

lib/src/query.dart
Buggaboo: I was thinking about [java's autocloseable](https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html) lately. It would be interesting to emulate this somehow.

lib/src/query.dart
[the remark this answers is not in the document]
vaind: It is supported in the C-API, it's `PropertyType::StringVector`

basically - it if you store a vector of strings in a property tagged this way, it can run some queries (I think right now it's only "contains"). It's a useful thing and should definitely be covered later (as in in the "all-type-support issue") in Dart as well

test/box_test.dart
```diff
@@ -92,12 +99,243 @@ void main() {
       expect(fetchedItems[1], equals(null));
       expect(fetchedItems[2].text, equals("Two"));
     });
+  });
+
+  group("query", () {
```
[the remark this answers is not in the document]
vaind: Specifically for queries, since there are so many different combinations, it would take too much time to use "standard" approach of creating a query, executing it and checking its result to cover 100 % query type/condition combinations (especially in Go, since there are currently 17 supported types...) 
In order to simplify it, you can define test-cases and just execute each of them using the same code. 
In it's most simple form a test case is a pair of `expectedQueryDescribeParamsResult` and `QueryCondition`. You pass this to a test "executor":
```dart
void testQuery(testCase) {
var query = box.query(testCase.conditions).build()
var actualDesc = query.describeParams()
expect(actualDesc, testCase.expectedDesc)
}
```

What Go query tests do on the line I've linked above does exactly that: call `testQueries()` with a huge list of test cases. They're a little more complex then just comparing description but that is actually the most important thing (as long as you test at least a couple of queries fully - which you already do.

lib/src/query.dart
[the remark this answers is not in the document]
Buggaboo: Done

README.md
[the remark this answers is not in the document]
Buggaboo: Removed.

lib/src/query.dart
Buggaboo: Done

lib/src/bindings/bindings.dart
```diff
@@ -106,8 +106,8 @@ class _ObjectBoxBindings {
   // query
   obx_query_t obx_query_create;
   obx_query_close_dart_t obx_query_close;
-  obx_query_find_t<Void, int> obx_query_find;
-  obx_query_find_t<Void, int> obx_query_find_ids; // ids = OBX_id_array, else OBX_bytes_array
+  obx_query_find_t<Uint64, int> obx_query_find;
+  obx_query_find_t<Uint64, int> obx_query_find_ids; // ids = OBX_id_array, else OBX_bytes_array
```
vaind: these types look rather weird - why are they typed the same?

lib/src/query.dart
vaind: yeah, that'd probably work, something like `with(arg1, arg2, arg3, fn(){})`...

lib/src/query.dart
```diff
@@ -0,0 +1,766 @@
...
+    ConditionOp   op   = condition._op;
+    int propertyId = qc._propertyId;
+
+    // TODO remove debug code
+    // print("type: ${type.toString()}, op: ${op.toString()}");
+
+    // do the typecasting here, we can't generalize to an op method
+    // due to the differing number of parameters per ConditionType
+    try {
+      switch (type) {
```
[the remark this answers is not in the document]
Buggaboo: Done.

lib/src/query.dart
Buggaboo: Done.

lib/src/query.dart
Buggaboo: I removed the flags.

lib/src/bindings/bindings.dart
Buggaboo: Good catch, I had a brainfart there. I thought they were going to be different types, being a `@ffi.Struct` and all that. But my experiment failed.

Buggaboo: Reverted.

lib/src/query.dart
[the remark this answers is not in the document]
Buggaboo: Done.

lib/src/query.dart
Buggaboo: Moved out to constants.

lib/src/annotations.dart
vaind: this would require further discussion so I'm not sure about merging it as part of this PR

bin/objectbox_model_generator/lib/src/code_chunks.dart
```diff
@@ -43,4 +46,59 @@ class CodeChunks {
...
+          fieldType = "Integer";
+          break;
+        case OBXPropertyType.Byte:
+          continue integer;
+        case OBXPropertyType.Short:
+          continue integer;
+        case OBXPropertyType.Char:
+          continue integer;
+        case OBXPropertyType.Long:
+          continue integer;
```
vaind: Doesn't this work in dart? 
```dart

        case OBXPropertyType.Int:
        case OBXPropertyType.Long:
        case OBXPropertyType.Short:
        case OBXPropertyType.Byte:
        case OBXPropertyType.Char:
          fieldType = "Integer";
```

lib/src/bindings/signatures.dart
```diff
@@ -55,3 +55,70 @@ typedef obx_box_put_native_t = Int32 Function(Pointer<Void> box, Uint64 id, Poin
...
+typedef obx_qb_order_dart_t   = int Function(Pointer<Void> builder, int property_id, int flags);
+
+// query
+
+typedef obx_query_t = Pointer<Void> Function(Pointer<Void> builder);
+
+typedef obx_query_close_native_t = Int32 Function(Pointer<Void> query);
+typedef obx_query_close_dart_t =   int Function(Pointer<Void> query);
+
+typedef obx_query_find_t<T> = Pointer<Uint64> Function(Pointer<Void> query, T offset, T limit);
```
vaind: this still doesn't seem right, I'd have expected something more similar to box_get_all (`	typedef obx_box_get_all_native_t = Pointer<Uint64> Function(Pointer<Void> box);`) and a separate method for find and find_ids (they have different return types

```suggestion
typedef obx_query_find_t = Pointer<Void> Function(Pointer<Void> query, Uint64 offset, Uint64 limit);
typedef obx_query_find_ids_t = Pointer<Uint64> Function(Pointer<Void> query, Uint64 offset, Uint64 limit);
```

but maybe I'm missing something

bin/objectbox_model_generator/lib/src/generator.dart
```diff
@@ -78,7 +78,8 @@ class EntityGenerator extends GeneratorForAnnotation<obx.Entity> {
             flags |= OBXPropertyFlag.ID;
             hasIdProperty = true;
           } else if (annotType == "Property") {
-            // nothing special
+            fieldType = annotVal.getField("type")?.toIntValue();
+            flags |= annotVal.getField("flag")?.toIntValue();
```
vaind: crashes if `flag` annotation is not present.
maybe?
```suggestion
            flags |= annotVal.getField("flag")?.toIntValue()?? 0;
```

lib/src/query/query.dart
```diff
@@ -0,0 +1,563 @@
...
+    return _op(ConditionOp.lt, p, null);
+  }
+
+  QueryCondition operator < (double p) => lessThan(p);
+  QueryCondition operator > (double p) => greaterThan(p);
+  QueryCondition operator == (double p) => equals(p);
+}
+
+class QueryBooleanProperty extends QueryProperty {
+  QueryBooleanProperty(int entityId, int propertyId) : super(entityId, propertyId, type);
```
vaind:
```suggestion
  QueryBooleanProperty(int entityId, int propertyId, int type) : super(entityId, propertyId, type);
```

lib/src/query.dart
Buggaboo: Done.

lib/src/query.dart
Buggaboo: Done.

bin/objectbox_model_generator/lib/src/code_chunks.dart
Buggaboo: Nope, no fall-through, try this:

```dart
// switch ...
  label meh:
  case a:
    return something();
  case b:
    continue a; // to simulate fall-thru
  case c:
     somethingElse();
     break;
// ...
```

I fixed a crap load of things, please try again.

lib/src/bindings/signatures.dart
Buggaboo: I didn't feel like repeating myself (DRY), if the generic function can switch out the `int` for `Uint64`.

bin/objectbox_model_generator/lib/src/generator.dart
Buggaboo: I fixed all of that, please check again.

lib/src/query/query.dart
Buggaboo: Sorry, about that oversight. I also found it.

README.md
```diff
@@ -95,6 +95,44 @@ print("refetched note: ${box.get(note.id)}");
...
+
+```dart
+// var store ...
+// var box ...
+
+box.putMany([Note(), Note(), Note()]);
+box.put(Note.construct("Hello world!"));
+
+final text = Note_.text;
+final queryNullText = box.query(text.isNull()).build();
```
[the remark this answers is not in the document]
Buggaboo: Done.

bin/objectbox_model_generator/lib/src/code_chunks.dart
```diff
@@ -43,4 +46,59 @@ class CodeChunks {
...
+        default:
+          throw InvalidGenerationSourceError("Unsupported property type (${f.type}): ${readEntity.name}.${name}");
+      }
+
+      ret.add("""
+        static final ${name}PropertyType = ${f.type};
+        static final ${name}PropertyId = ${f.id.id};
+        static final ${name} = Query${fieldType}Property(entityId, ${name}PropertyId, ${name}PropertyType);
```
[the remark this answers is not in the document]
Buggaboo: Done.

lib/src/query/query.dart
```diff
@@ -0,0 +1,571 @@
...
+        continue integer;
+      case OBXPropertyType.Byte:
+        continue integer;
+      case OBXPropertyType.Short:
+        continue integer;
+      case OBXPropertyType.Char:
+        continue integer;
+      case OBXPropertyType.Bool:
+        continue integer;
+    }
```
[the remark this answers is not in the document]
Buggaboo: I believe you made the change already, can you `resolve` this issue?

lib/src/query/query.dart
```diff
@@ -0,0 +1,571 @@
...
+}
+
+/**
+ * The first element of the chain
+ * contains the enum representation of
+ * the to-be-constructed query builder.
+ * This design allows nested chains inside
+ * the chain.
+ */
+class QueryCondition {
```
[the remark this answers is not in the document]
Buggaboo: I believe that `any`/`all` conditions should be group together as many conditions as possible, if they're on the same logical level anyway.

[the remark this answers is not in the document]
Buggaboo: Hm, I have to learn to read better. For single conditions, this makes sense. I'll take a look again.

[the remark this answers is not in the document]
Buggaboo: I prefer composition to inheritance in this case, because the all/any operators just take a builder and an array of conditions. No `entityId`, no `propertyId`.

test/query_test.dart
```diff
@@ -0,0 +1,325 @@
...
+
+void main() {
+  Store store;
+  Box box;
+
+  group("query", () {
+    setUp(() {
+      store = Store([TestEntity_OBXDefs]);
+      box = Box<TestEntity>(store);
+    });
```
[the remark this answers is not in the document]
Buggaboo: For some reason, since the yaml update, the `dylibs` are broken on my machine. I can't run or test anything that's related to objectbox.

lib/src/query/query.dart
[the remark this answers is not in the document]
Buggaboo: Are you implying that you want to get rid of `build` and directly create the condition objects? I believe that the _swift_ project does that, they just pass a closure.

test/box_test.dart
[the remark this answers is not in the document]
vaind: I've noticed you added these tests, thanks.

lib/src/query/query.dart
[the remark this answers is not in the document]
Buggaboo: As long as the algorithm efficiently creates trees of conditions, I'm fine with it. You've collapsed some classes together, it looks more cohesive like this.

I especially like this:
```dart
final list = [this, ...rest];
```
I wasn't aware of this syntax.

Merged.

lib/src/bindings/bindings.dart
```diff
@@ -126,7 +126,7 @@ class _ObjectBoxBindings {
     if (Platform.isWindows)
       libName += ".dll";
     else if (Platform.isMacOS)
-      libName = "lib" + libName + ".dylib";
+      libName = "lib/lib" + libName + ".dylib";
```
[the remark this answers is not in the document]
Buggaboo: I got this, by moving the lib one level lower (project root):

```bash
...
 Invalid argument(s): Failed to load dynamic library (dlopen(libobjectbox.dylib, 1): no suitable image found.  Did find:
        libobjectbox.dylib: code signature in (libobjectbox.dylib) not valid for use in process using Library Validation: mapped file has no cdhash, completely unsigned? Code has to be at least ad-hoc signed.)
  dart:ffi                                             new DynamicLibrary.open
...
```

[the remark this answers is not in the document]
vaind: whoa, interesting... I'll try it on mac myself to make sure I can reproduce.Can you open a separate ticket though? I'm going to merge this PR and this is unrelated and not yet clear how we actually want to solve it (so it doesn't break for other people).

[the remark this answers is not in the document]
Buggaboo: I reverted the change already. I just made a note of it on the README.

vaind:
I've merged this, doing a couple of small changes in the process
* reverted the double equals() for now to prevent users assuming it actually works 100 % correctly
* added tests to make sure conditions are combined together as expected
* formatted
* updated the generator test case

Buggaboo:
Omg! Let's do a beer some time to celebrate.

Issue #36, opened 2019-10-15
Box provide missing methods


vaind:
Box already has get/put as well as their bulk variants. To be more up to par with the rest of the bindings, the following methods should be implemented:

* [x] count(optional limit)
* [x] contains(id)
* [x] isEmpty()
* [x] remove(id)
* [x] removeAll()

See [box.go](https://github.com/objectbox/objectbox-go/blob/master/objectbox/box.go) and [go docs](https://godoc.org/github.com/objectbox/objectbox-go/objectbox#Box) for reference

vaind:
Thank you @liquidiert :+1: I've merged the PR after resolving some issues caused by the huge query PR changelist.

Could you maybe have a look at the last method (or two): `contains(id), containsMany(ids)` since you're already familiar with the code and its tests. If so, please start it on a new branch based on the latest `dev` to prevent merge conflicts, as there have been major changes today.

liquidiert:
Of course :+1: I'm happy to help 😄

vaind:
thanks for implementing this @liquidiert :tada:

PR #42, opened 2019-10-20
implemented .count(), .isEmpty(), .removeAll()

liquidiert:
This PR addresses Issue #36.  
- implemented .count() and .countMax() functions from C api -> created new bindings
- implemented .isEmpty() function from C api -> created new binding
- implemented .removeAll() [remove all objects in box, not all ids from iterable -> still needs to be implemented?] function from C api -> created new binding

### DIFF
--- a/lib/src/box.dart
+++ b/lib/src/box.dart
@@ -166,5 +166,29 @@ class Box<T> {
     return _getMany(() => checkObxPtr(bindings.obx_box_get_all(_cBox), "failed to get all objects from box", true));
   }
 
+  // same as calling maxCount with limit := 0
+  int count() {
+    return maxCount(limit: 0);
+  }
+
+  int maxCount({int limit}) {
+    Pointer<Uint64> _count = Pointer<Uint64>.allocate();
+    checkObx(bindings.obx_box_count(_cBox, limit, _count));
+    return _count.load<int>();
+  }
+
+  bool isEmpty() {
+    Pointer<Uint8> _isEmpty = Pointer<Uint8>.allocate();
+    checkObx(bindings.obx_box_is_empty(_cBox, _isEmpty));
+    return _isEmpty.load<int>() > 0 ? true : false;
+  }
+
+  int removeAll() {
+    Pointer<Uint64> _removedItems = Pointer<Uint64>.allocate();
+    checkObx(bindings.obx_box_remove_all(_cBox, _removedItems));
+    return _removedItems.load<int>();
+  }
+
   get ptr => _cBox;
+
 }
--- a/lib/src/box.dart
+++ b/lib/src/box.dart
@@ -1,5 +1,7 @@
 import "dart:ffi";
 
+import 'package:objectbox/src/common.dart';
+
 import "store.dart";
 import "bindings/bindings.dart";
 import "bindings/constants.dart";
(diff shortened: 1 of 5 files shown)


liquidiert:
Hey guys, it's my first PR ever, so please don't destroy it that much 😄  also advice is much appreciated :)

lib/src/box.dart
```diff
@@ -166,5 +166,29 @@ class Box<T> {
...
+  // same as calling maxCount with limit := 0
+  int count() {
+    return maxCount(limit: 0);
+  }
+
+  int maxCount({int limit}) {
+    Pointer<Uint64> _count = Pointer<Uint64>.allocate();
+    checkObx(bindings.obx_box_count(_cBox, limit, _count));
+    return _count.load<int>();
+  }
```
vaind: Having two methods is not necessary. I assume you've been inspired by the Go implementation, however, it needs to use two methods because there are no optional arguments in Go.

In dart, we can have just `count({int limit})`

lib/src/box.dart
```diff
@@ -166,5 +166,29 @@ class Box<T> {
...
+
+  bool isEmpty() {
+    Pointer<Uint8> _isEmpty = Pointer<Uint8>.allocate();
+    checkObx(bindings.obx_box_is_empty(_cBox, _isEmpty));
+    return _isEmpty.load<int>() > 0 ? true : false;
```
vaind: `Pointer<T>` needs manual freeing. You can have a look at other code in `Box` to see how it's used.

In this example, it would look sth. like:
```dart
bool isEmpty() {
    Pointer<Uint8> _isEmpty = Pointer<Uint8>.allocate();
    try {
        checkObx(bindings.obx_box_is_empty(_cBox, _isEmpty));
        return _isEmpty.load<int>() > 0 ? true : false;
    } finally {
        _isEmpty.free();
    }
}
```

test/box_test.dart
```diff
@@ -93,6 +93,18 @@ void main() {
...
+      final List<int> ids = box.putMany(items);
+      
+      int count = box.count();
+      expect(count, greaterThanOrEqualTo(0));
+      bool isEmpty = box.isEmpty();
+      expect(isEmpty, equals(false));
+      int removed = box.removeAll();
+      expect(removed, greaterThan(0));
+    });
+
```
vaind: Could you split this up to separate tests (per function)? Also, it should be a little more thorough, these are fairly important methods :)

e.g. sth like the following for `count()`:
```dart
    test("count", () {
      expect(box.count(), equals(0));
      final List<TestEntity> items = ["One", "Two"].map((s) => TestEntity.construct(s)).toList();
      box.putMany(items);
      
      expect(box.count(), equals(2));
    });
```

lib/src/box.dart
liquidiert: Ah yeah of course!

liquidiert: Do you think a named parameter or a true optional `[limit = 0]` param should be used? As the docs mention .count() can be called with an optional limit -> which implies the parameter doesn't need to be named.

vaind: I think named parameter would be good in this case

liquidiert:
I have two questions regarding the .remove() method:
- shall it work only for single ids or also for iterables of ids?
- and if you remove an id that has been removed before ObjectBox returns a 404 exception -> do we need to handle that exception or is it intended that way?
thanks for the fast answers btw :)

vaind:
> I have two questions regarding the .remove() method:
> 
> * shall it work only for a single id or also for iterables of ids?

Thanks for a reminder. We have `RemoveMany()` in Go. How would you implement `box.remove()` for two types while keeping it statically typed? labeled parameters? Maybe it makes sense to have a speparate method in Dart as well, since `obx_box_remove_many()` returns a count of removed IDs (as opposed to failing when removing a single one).


> * and if you remove an id that has been removed before, ObjectBox returns a 404 exception -> do we need to handle that exception or is it intended that way?

What would be an idiomatic way to handle this in Dart?

>   thanks for the fast answers btw :)
you're welcome :)

liquidiert:
I would've implemented it with type `dynamic` and a type check, but if we already have a different method that's probably better

liquidiert:
I don't know either unfortunately :/ that's why I've asked but I at least could wrap it with a more meaningful exception message than `ObjecBoxException 404`

vaind:
> I don't know either unfortunately :/ that's why I've asked but I at least could wrap it with a more meaningful exception message than `ObjecBoxException 404`

In that case, let's change to bool result and return false for this specific exception (not-found). Other exceptions, if any, should still be propagated.

liquidiert:
> I would've implemented it with type `dynamic` and a type check, but if we already have a different method that's probably better

so are we going with the two methods approach?

liquidiert:
Ok rethought about the approaches and went with two methods :+1:

PR #45, opened 2019-10-21
implemented .contains/Many(); test fails for query.describe()

liquidiert:
implemented .contains(), ,containsMany() and their corresponding unit tests

### DIFF
--- a/lib/src/box.dart
+++ b/lib/src/box.dart
@@ -186,6 +186,28 @@ class Box<T> {
     }
   }
 
+  bool contains(int id) {
+    Pointer<Uint8> _contains = Pointer<Uint8>.allocate();
+    try {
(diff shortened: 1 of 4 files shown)


lib/src/box.dart
```diff
@@ -186,6 +186,28 @@ class Box<T> {
     }
   }
 
+  bool contains(int id) {
+    Pointer<Uint8> _contains = Pointer<Uint8>.allocate();
```
vaind: I've forgotten to bring this up yesterday. Could you please not use an underscore for local variables? It may be confused with private instance vars, especially in larger methods

liquidiert: Yeah sorry, "fixed" that :+1:

vaind:
thanks :ice_cream:

liquidiert:
You're welcome 😄

Issue #48, opened 2019-10-22
Property query


Buggaboo:
[Another chunk of code](https://github.com/objectbox/objectbox-c/blob/master/include/objectbox.h#L925-L1008) coming this way, stay tuned.

vaind:
I think the original PR was in a pretty good shape so feel free to open a new one with the dart 2.6 support when it works for you.

greenrobot-team:
This was added with #108.

PR #51, opened 2019-10-23
WIP - Property query

Buggaboo:
#48 I need help with the tests, and some feedback wrt to the design.

My count test is broken. What am I doing wrong?

Please, don't merge yet.


vaind:
> My count test is broken. What am I doing wrong?

Couldn't find anything... current dart v2.5.2 still fails to properly show capture the exception (issue #1)  and the lib doesn't compile with the latest dev v2.6 which should have the issue resolved and show you the proper error. I'll try to have a look at the errors given by 2.6 tomorrow and let you know.

vaind:
Regarding the design, I'll have a better look tomorrow. Just one thing that occurred to me would be nice:
```
final queryInt = query.property(tLong); // lose the type in the method name
int sum = queryInt.sum();
```
Maybe we could handle it with the type system somehow - using generics on `QueryProperty` and getting the generic type out of it in the declaration of `property()` method

Buggaboo:
> Regarding the design, I'll have a better look tomorrow. Just one thing that occurred to me would be nice:
> 
> ```
> final queryInt = query.property(tLong); // lose the type in the method name
> int sum = queryInt.sum();
> ```
> 
> Maybe we could handle it with the type system somehow - using generics on `QueryProperty` and getting the generic type out of it in the declaration of `property()` method

Done.

Buggaboo:
> > My count test is broken. What am I doing wrong?
> 
> Couldn't find anything... current dart v2.5.2 still fails to properly show capture the exception (issue #1) and the lib doesn't compile with the latest dev v2.6 which should have the issue resolved and show you the proper error. I'll try to have a look at the errors given by 2.6 tomorrow and let you know.

It might be a bug wrt to signedness (e.g. UintX vs IntX), but I'm not sure.

vaind:
With all the errors you're getting, it seems something is fundamentally wrong. I'd maybe wait for 2.6 before spending too much time trying to debug it. With a proper error message, it should be straightforward.

Buggaboo:
> With all the errors you're getting, it seems something is fundamentally wrong. I'd maybe wait for 2.6 before spending too much time trying to debug it. With a proper error message, it should be straightforward.

Ugh. Can you please start a *dev-2.6* branch? I'll start hacking that instead. This is driving me crazy.

Buggaboo:
I retraced my steps and reimplemented the cstructs used in the `find` calls without `Struct`, like we used to with `OBX_array_ids`. And I still have the same error. Something is fundamentally broken.

Buggaboo:
Since 2.6 is the new stable version. I'm closing this one.

Issue #54, opened 2019-10-24
Switch to Dart v2.6


vaind:
Dart v2.6 is going to break compatibility in FFI (it was in "preview" after all, so no surprise there). From what I've seen with the current dev version (.8), it was rather significant so I have a feeling we'll have to drop Dart v2.5 compatibility to be able to switch to v2.6

See [the changes](https://github.com/dart-lang/sdk/blob/master/CHANGELOG.md)  that have already landed in the dev release and [the FFI project board](https://github.com/dart-lang/sdk/projects/13) for what's coming.

liquidiert:
Memory management has been removed 😍

greenrobot:
There's no definite timeline yet on the 2.6 release, I suppose?

How about

1. When we get really close to a 2.6 release, e.g. RC2 or final release. We do a 2.5 release, make the adjustments for 2.6 and release this asap. Afterwards we focus on 2.6 only.
2. Like 1.), but prep a 2.6 branch in parallel

This way, we have another 2.5 release.

vaind:
Some volunteers who'd like to pitch in with the 2.6 support?  @liquidiert or @Buggaboo maybe? :smile:

Buggaboo:
Sure, I'd like to get property query out of the way first.

liquidiert:
Would be happy to help 😁

Buggaboo:
> Sure, I'd like to get property query out of the way first.

I changed my mind, I'm stuck. @vaind Can you open a new `dev-2.6` branch? Dakujeme.

We should start with the imports.

I expect `free` and `allocate` to break. Probably `Struct`, anything related to *ffi* really.
Let's see if `attachFinalizer` or something like that is available, so we can `deallocate`, or by whatever `free` has been replaced with. 

I'm reluctant to use brew's dart-2.6-devX builds; I prefer to wait for the non-dev, _stable_(?) 2.6 version. I can't be arsed to compile my own either.

vaind:
> ... Dakujeme.

`dev-dart26` You're welcome :rofl: 

> I'm reluctant to use brew's dart-2.6-devX builds; I prefer to wait for the non-dev, stable(?) 2.6 version. I can't be arsed to compile my own either.

I just download and extract the SDK & set its path in the IDE. Maybe that works for you as well: https://dart.dev/tools/sdk/archive#dev-channel

Buggaboo:
Done.

liquidiert:
[...]

PR #57, opened 2019-10-25
WIP - support for dart26

Buggaboo:
Relevant changes since 2.6
---------------------------
* Changed `Struct<T>` to `Struct`
* To dereference a Struct to get to the values you need to `ref`
* `free` and `allocate` are part of `package:ffi` as independent functions
* `load` and `store` became a getter/setter called `value`
* `elementAt` has indexing syntax-sugar i.e. `ptr[i]`
* asExternalType bla, is [replaced with `asTypedList`](https://github.com/dart-lang/sdk/commit/91c6dc15de164e69052de7c6b1d608c57e5dea7d) to create dart native list types

Update
-------
* I'm not 100% sure that `free(someUtf8Instance)` does the job, since `Utf8 extends Struct` and does *not* clearly [address the `Uint8` allocation](https://github.com/dart-lang/ffi/blob/master/lib/src/utf8.dart#L56).
* To move forward, I think we should reimplement all ByteBuffer related stuff as proper dart Structs.
* We should check if `attachFinalizer` is available, and hook on `free` from there. (TODO anyone else other than me)

At least everything generates, compiles, and the coverage is back to 100%.

This addresses Issue #54

Sources:
* [package:ffi allocate free](https://github.com/dart-lang/ffi/commit/f46c1f42c6f7f1938f2ff27c573da72d57c47ded#diff-356c02ce44d5d1bab5753b24a1d483e6)
* [ffi samples](https://github.com/dart-lang/samples/tree/master/ffi)
* [this load/store related commit](https://github.com/dart-lang/sdk/commit/597cd06aec7f64b3defd482e84f5d75bef15705b)


nalenz-objectbox:
- FFI 0.1.3 has been released, so [this line](https://github.com/Buggaboo/objectbox-dart/blob/dev-dart26/pubspec.yaml#L13) can be changed to `ffi: ^0.1.3`
- to improve our [analytics page](https://pub.dev/packages/objectbox#-analysis-tab-), I think it's best to check the arising warnings using [pana](https://github.com/dart-lang/pana) beforehand, maybe even before pushing
- unfortunately, it seems like the [contribution guide](https://github.com/Buggaboo/objectbox-dart/blob/dev-dart26/CONTRIBUTING.md#code-contributions) needs to be updated in that case, because pana only accepts code formatted with `dartfmt`, not `dartfmt -l 120`
- Dart 2.6 has been released, so we can use `sdk: ">=2.6.0 <3.0.0"` as [environment dependency](https://github.com/Buggaboo/objectbox-dart/blob/dev-dart26/pubspec.yaml#L9). This needs to be changed [in the generator](https://github.com/Buggaboo/objectbox-dart/blob/dev-dart26/generator/pubspec.yaml#L9) as well
- at least during testing and pana checks, I guess also use `objectbox: path: ..` instead of the `objectbox: 0.4.0` pub.dev dependency for objectbox_generator?
- the code can be cleaned from a lot of unnecessary imports and other hints and (linter) warnings, as reported by pana
- in `bindings.dart`, the following error is reported for [lines 157 and 158](https://github.com/Buggaboo/objectbox-dart/blob/dev-dart26/lib/src/bindings/bindings.dart#L157): `The type argument for the pointer, 'NativeFunction<Void Function(Pointer<OBX_bytes_array>)>' must be a 'NativeFunction' in order to use 'asFunction'. Try changing the type argument to be a 'NativeFunction'.`

vaind:
> * unfortunately, it seems like the [contribution guide](https://github.com/Buggaboo/objectbox-dart/blob/dev-dart26/CONTRIBUTING.md#code-contributions) needs to be updated in that case, because pana only accepts code formatted with `dartfmt`, not `dartfmt -l 120`

pana takes the `-l` argument as well. While it will still report on pub.dev, these suggestions have no score impact.

> * at least during testing and pana checks, I guess also use `objectbox: path: ..` instead of the `objectbox: 0.4.0` pub.dev dependency for objectbox_generator?

Hmm, what does that help with?

nalenz-objectbox:
> Hmm, what does that help with?
Actually nothing really, just shortly during development right now, because the current regular `objectbox` pub.dev package still contains the Dart `<=2.5.0` restriction

PR #75, opened 2019-11-20
Help required: query property

Buggaboo:
Remaining issues (2020 May 25)
---------------------------------
1. For example,`.find(replaceNullWith:-2)` on _tLong_ works, but still not on the other integerish types.
2. Doubles could be initialized to null (expected), but floats are set to 0.0f by default, thus making _replaceWithNull_ unnecessary. Same thing as last year.
3. Same syntax issue, we haven't addressed that at all. My gut feeling says we should apply extensions, and drop the generics.

So... I'm still stuck 6 months later. Ideas?

Remaining issues (2019 Nov 29)
---------------------------
1. Signed negative values are returned as unsigned values by the `find` function on _tShort_ and _tInteger_ properties, e.g. [`find(replaceNullWith: -2)`](https://github.com/Buggaboo/objectbox-dart/blob/query-property-dev-dart26/test/query_property_test.dart#L341)
2. Float properties are never null, [so regardless of the `replaceNullWith` parameter, uninitialized values will always return `0.0`. ](https://github.com/Buggaboo/objectbox-dart/blob/query-property-dev-dart26/test/query_property_test.dart#L357)
3. Casting to PropertyQuery is an unnecessary pain.

Right now we have:
```dart
final query = box.query(tInteger.greaterThan(0)).build();
final qp = query.property(tInteger);
```
Where `tInteger` is referenced to return the concrete `IntegerPropertyQuery`, but the return type is still the limiting factor, i.e. the compiler will make it `PropertyQuery` again, because the generic param `<PQ extends PropertyQuery>` requirement applies.

We could change the syntax:

```dart
final q = box.query(tInteger.greaterThan(0)).build();
final qp = (tInteger.query(q)..distinct = true).average(); // Entity_.tInteger.query function is generated
q.close(); qp.close();
```


lib/src/bindings/signatures.dart
```diff
@@ -152,6 +152,14 @@ typedef obx_query_visit_native_t = Int32 Function(
 typedef obx_query_visit_dart_t = int Function(
     Pointer<Void> query, Pointer<Void> visitor, Pointer<Void> user_data, int offset, int limit);
 
+// query property
+typedef obx_query_prop_t<T> = Pointer<Void> Function(Pointer<Void> query, T propertyId); // Uint32 -> int
+typedef obx_query_prop_close_t<T> = T Function(Pointer<Void> query); // obx_err -> Int32 -> int
+typedef obx_query_prop_find_t<T, V> = Pointer<T> Function(Pointer<Void> query, Pointer<V> value_if_null);
```
vaind:
```suggestion
typedef obx_query_prop_find_t<T, V> = Pointer<T> Function(Pointer<Void> query, Pointer<V> value_if_null);
typedef obx_query_prop_find_native_t<T extends NativeType, V extends NativeType> = T Function(Pointer<Void> query, Pointer<V> value_if_null); // no Pointer<T>, code analysis fails on usage
```

lib/src/bindings/bindings.dart
```diff
@@ -284,6 +308,31 @@ class _ObjectBoxBindings {
...
+    obx_query_prop_avg = _fn<obx_query_prop_op_t<Int32, Double>>("obx_query_prop_avg").asFunction();
+    obx_query_prop_min = _fn<obx_query_prop_op_t<Int32, Double>>("obx_query_prop_min").asFunction();
+    obx_query_prop_max = _fn<obx_query_prop_op_t<Int32, Double>>("obx_query_prop_max").asFunction();
+    obx_query_prop_sum = _fn<obx_query_prop_op_t<Int32, Double>>("obx_query_prop_sum").asFunction();
+
+    obx_query_prop_min_int = _fn<obx_query_prop_op_t<Int32, Int64>>("obx_query_prop_min_int").asFunction();
+    obx_query_prop_max_int = _fn<obx_query_prop_op_t<Int32, Int64>>("obx_query_prop_max_int").asFunction();
+    obx_query_prop_sum_int = _fn<obx_query_prop_op_t<Int32, Int64>>("obx_query_prop_sum_int").asFunction();
+
+    obx_query_prop_string_find = _fn<obx_query_prop_find_t<OBX_string_array, Int8>>("obx_query_prop_string_find").asFunction();
```
vaind: to get rid of code analysis reports (the same for the remaining finds())
```suggestion
    obx_query_prop_string_find = _fn<obx_query_prop_find_native_t<Pointer<OBX_string_array>, Int8>>("obx_query_prop_string_find").asFunction();
```

lib/src/bindings/signatures.dart
```diff
@@ -152,6 +152,14 @@ typedef obx_query_visit_native_t = Int32 Function(
...
+typedef obx_query_prop_distinct_t<T, V> = T Function(Pointer<Void> query, V distinct); // T = (Int32, int), V = (Int8, int)
+typedef obx_query_prop_distinct2_t<T, V> = T Function(Pointer<Void> query, V distinct, V caseSensitive);
+typedef obx_query_prop_op_t<T, V> = T Function(Pointer<Void> query, Pointer<V> out);
```
vaind:
```suggestion
typedef obx_query_prop_op_t<T, V extends NativeType> = T Function(Pointer<Void> query, Pointer<V> out);
```

test/query_property_test.dart
```diff
@@ -0,0 +1,406 @@
...
+  test(".find() default values on null results" , () {
+    box.putMany(integerList);
+    box.putMany(stringList);
+    box.putMany(floatList);
+
+    final queryIntegers = box.query(tLong.greaterThan(1000)).build();
+    final queryFloats   = box.query(tDouble.greaterThan(1000.0)).build();
+    final queryStrings  = box.query(tString.equals("can't find this")).build();
+
+    final integerDefaultValue = -2;
```
vaind: this is misunderstood, the default value will be used in place of `NULL` on a given property, i.e you need to first insert objects with NULLs.

lib/src/bindings/bindings.dart
```diff
@@ -284,6 +308,31 @@ class _ObjectBoxBindings {
...
     obx_query_visit = _fn<obx_query_visit_native_t>("obx_query_visit").asFunction();
 
+    // query property
+    obx_query_prop = _fn<obx_query_prop_t<Uint32>>("obx_query_prop").asFunction();
+    obx_query_prop_close = _fn<obx_query_prop_close_t<Int32>>("obx_query_prop_close").asFunction();
+
+    obx_query_prop_distinct =        _fn<obx_query_prop_distinct_t<Int32, Int8>>("obx_query_prop_distinct").asFunction();
+    obx_query_prop_distinct_string = _fn<obx_query_prop_distinct2_t<Int32, Int8>>("obx_query_prop_distinct_string").asFunction();
+
+    obx_query_prop_count = _fn<obx_query_prop_op_t<Int32, Uint64>>("obx_query_prop_string_find").asFunction();
```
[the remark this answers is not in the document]
Buggaboo: Doh. Thanks for finding it.

test/query_property_test.dart
Buggaboo: Ah, thanks for clarifying that.

Buggaboo: I updated the param name and tests to reflect that.

greenrobot-team:
@Buggaboo I'm trying to finish this in #108, feedback is still welcome.

greenrobot-team:
Merged as part of #108.

PR #78, opened 2019-11-28
iOS support

### LINKED ISSUE #47: iOS support
After the [flutter mobile example](https://github.com/objectbox/objectbox-dart/issues/8) is prepared, it can be used to try if we can get the library to work on iOS. Seems like a `.framework` would be necessary. Hopefully, we can extract it from an [objectbox-swift release](https://github.com/objectbox/objectbox-swift/releases)

vaind:
Brings iOS support (tested on example/flutter/objectbox_demo). Fixes #47 

~~Currently doesn't work properly because it requires objectbox-swift sourced `ObjectBox.framework` to be extracted in the ios folder to build correctly. Depends on https://github.com/flutter/flutter/issues/45778~~

While the issue is still open, we can circumvent it by packaging the `ObjectBox.framework` directly when publishing to pub.dev. That's what this PR currently does.


ios/download-framework.sh
```diff
@@ -0,0 +1,21 @@
...
+obxSwiftVersion="1.2.0"
+
+dir=$(dirname "$0")
+
+url="https://github.com/objectbox/objectbox-swift/releases/download/v${obxSwiftVersion}/ObjectBox-framework-${obxSwiftVersion}.zip"
+zip="${dir}/fw.zip"
+
+curl --location --fail --output "${zip}" "${url}"
+
+rm -rf "${dir}/Carthage"
```
greenrobot: Danger zone? E.g. what if we are not the only ones here?

ios/.gitignore
```diff
@@ -0,0 +1,3 @@
+# NOTE: comment out before publishing - the binaries need to be uploaded
```
greenrobot: I feel like some of the process here should be documented somewhere?

vaind: all "before publishing" are covered by the release notes.

I've also added `ios/README.md`, moving info from the podspec and expanding upon it to make the dev's life easier.

ios/download-framework.sh
vaind: the info was in the podspec, moved to ios/readme

vaind: no danger of removing anything unintentionally, this is run by a ObjectBox-Dart publisher before a release

PR #79, opened 2019-11-29
Android setup simplification

### LINKED ISSUE #73: Avoid android users specifying `objectbox-android` dependency
Currently, to use `objectbox-dart` with flutter on Android, a user must manually specify a dependency on `objectbox-android`, which is a gradle artifact published by the `objectbox-java` project. Ideally, we should specify this dependency here in the dart/flutter lib, so the consumers wouldn't have to do it.

Follow up from #2 
See my question here: https://github.com/flutter/flutter/issues/33227#issuecomment-547979929
A branch where I tried something (it didn't work though): https://github.com/objectbox/objectbox-dart/tree/flutter-plugin

vaind:
Fixes #73

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,13 +28,6 @@ dev_dependencies:
 Proceed based on whether you're developing a Flutter app or a standalone dart program:
 1. **Flutter** only steps:
     * Install the packages `flutter pub get`
-    * Add `objectbox-android` dependency to your `android/app/build.gradle` 
-        ```
-        dependencies {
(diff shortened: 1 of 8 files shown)


greenrobot-team:
Reviewing...

liquidiert:
@greenrobot-team dunno if it helps but I'm using object box 0.4.0 in one of my current projects and I can build my g parts 👌

greenrobot-team:
@liquidiert Thanks for that! Turns out it failed due to a regression introduced with an unrelated change in the development version.

Issue #52, opened 2019-10-24
Provide examples of Isolates i.e. don't do everything on the main thread


Buggaboo:
TODO
------
* free datastructures when exiting
* message passing, e.g. [ReceivePort](https://api.dartlang.org/stable/2.5.2/dart-isolate/ReceivePort-class.html), [SendPort](https://api.dartlang.org/stable/2.5.2/dart-isolate/SendPort-class.html)
* transactions, atomicity guarantee etc.

vaind:
Also, we should have test cases covering this usecase to make sure the native interactions work as expected (e.g. transactions/locking)

greenrobot:
For me as a Dart noob, it would be really interesting to see a minimal Isolate example. E.g. can we pass a Box into another Isolate? If so, we should be able to get some data objects there and copy those over into the main Isolate?

Buggaboo:
[I was playing with isolates here](https://github.com/Buggaboo/objectbox-dart/commit/561553f9dfb66116e408cd69125eaf641c9636f5)

It's not possible to pass around boxes, stores, or anything related to ffi.

The isolate can receive TestEntity objects, but vice versa the main terminates before receiving a TestEntity object from the isolate. I used to solve this with a completer, but since the isolate is in its own world...

vaind:
OK, so the [observers issue](https://github.com/objectbox/objectbox-dart/issues/142) prompted me to look into this...

Yes, multiple isolates can be used, the changes necessary are very simple: just allow creating Store from an existing C-API `OBX_Store` pointer. I've done so in this PR draft: #145

vaind:
implemented by #145

PR #95, opened 2020-03-04
Data visitor

### LINKED ISSUE #94: Query error for property with large String
I saved large string in database but when try to make query to get it I get error:
```
E/flutter (31707): [ERROR:flutter/lib/ui/ui_dart_state.cc(157)] Unhandled Exception: Exception: ObjectBoxException: find: 10098 Do not use vector-based find on 32 bit systems with big objects
E/flutter (31707): #0      checkObxPtr (package:objectbox/src/bindings/helpers.dart:15:5)
E/flutter (31707): #1      Query.find.<anonymous closure> (package:objectbox/src/query/query.dart:574:26)
E/flutter (31707): #2      Store.runInTransaction (package:objectbox/src/store.dart:69:16)
E/flutter (31707): #3      Query.find (package:objectbox/src/query/query.dart:573:19)
```
This problem is only for x86 Android Emulator and it is ok for physical devices.
I founded very similar issue for Java Objectbox that was fixed 1 year ago - [java issue](https://github.com/objectbox/objectbox-java/issues/669)

Thanks!

vaind:
Fixes #94

### DIFF
--- a/lib/src/bindings/data_visitor.dart
+++ b/lib/src/bindings/data_visitor.dart
@@ -2,17 +2,16 @@ import 'dart:ffi';
 import 'signatures.dart';
 import "package:ffi/ffi.dart" show allocate, free;
 
-typedef bool dataVisitorCallback(Pointer<Uint8> dataPtr, int length);
+int _lastId = 0;
+final _callbacks = <int, bool Function(Pointer<Uint8> dataPtr, int length)>{};
 
-int visitorId = 0;
-final callbacks = <int, dataVisitorCallback>{};
-
+// called from C, forwards calls to the actual callback registered at the given ID
 int _forwarder(Pointer<Void> callbackId, Pointer<Uint8> dataPtr, int size) {
-  if (callbackId == null) {
-    throw Exception("Data-visitor callback issued with NULL user_data");
+  if (callbackId == null || callbackId.address == 0) {
+    throw Exception("Data-visitor callback issued with NULL user_data (callback ID)");
   }
 
-  return callbacks[callbackId.cast<Int64>().value](dataPtr, size) ? 1 : 0;
+  return _callbacks[callbackId.cast<Int64>().value](dataPtr, size) ? 1 : 0;
 }
 
 /// A data visitor wrapper/forwarder to be used where obx_data_visitor is expected.
@@ -24,20 +23,20 @@ class DataVisitor {
 
   Pointer<Void> get userData => _idPtr.cast<Void>();
 
-  DataVisitor(dataVisitorCallback callback) {
+  DataVisitor(bool Function(Pointer<Uint8> dataPtr, int length) callback) {
     // cycle through ids until we find an empty slot
-    visitorId++;
-    var initialId = visitorId;
-    while (callbacks.containsKey(visitorId)) {
-      visitorId++;
+    _lastId++;
+    var initialId = _lastId;
+    while (_callbacks.containsKey(_lastId)) {
+      _lastId++;
 
-      if (initialId == visitorId) {
+      if (initialId == _lastId) {
         throw Exception("Data-visitor callbacks queue full - can't allocate another");
       }
     }
     // register the visitor
-    _id = visitorId;
-    callbacks[_id] = callback;
+    _id = _lastId;
+    _callbacks[_id] = callback;
 
     _idPtr = allocate<Int64>();
     _idPtr.value = _id;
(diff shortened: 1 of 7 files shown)


lib/src/bindings/data_visitor.dart
```diff
@@ -0,0 +1,50 @@
+import 'dart:ffi';
+import 'signatures.dart';
+import "package:ffi/ffi.dart" show allocate, free;
+
+int _lastId = 0;
+final _callbacks = <int, bool Function(Pointer<Uint8> dataPtr, int length)>{};
```
greenrobot: What do we need that for? E.g. this is one of the visitor C APIs:

obx_err obx_query_visit(OBX_query* query, obx_data_visitor* visitor, void* user_data, uint64_t offset, uint64_t limit);

I assume you want to pass "id" as `user_data`? Is this more "convenient" that passing in a lambda-like thingy as `visitor`? The latter would already be thread-safe, which the current map-based is not, I think?

lib/src/box.dart
```diff
@@ -160,34 +162,63 @@ class Box<T> {
...
     return _store.runInTransaction(TxMode.Read, () {
-      final bytesArray = cCall();
+      if (_supportsBytesArrays) {
+        final bytesArray = cGetArray();
+        try {
+          return _fbManager.unmarshalArray(bytesArray, allowMissing: allowMissing);
+        } finally {
+          bindings.obx_bytes_array_free(bytesArray);
+        }
+      }
```
greenrobot: Can we have an `} else {` here to better mark the two approaches?

lib/src/bindings/data_visitor.dart
vaind: Lambdas can't be passed as a C callback, FFI requires static functions. I've tried that (lambda) approach but had to resolve to this trampoline one instead (basically the same thing we do in Go)

lib/src/box.dart
vaind: sure

lib/src/bindings/data_visitor.dart
greenrobot: Sure, please make sure to document that in the code so that this explanation is not lost

greenrobot: > The latter would already be thread-safe, which the current map-based is not, I think?

I think this was not addressed yet. Does that require additional thread synchronization?

vaind: There is no concurrent memory access in Dart. All code in a single isolate executes on a single event-loop and isolates don't share memory.

greenrobot: Interesting concept. Didn't dig into into, but still have a question. We don't have just Dart but also native code. What happens when you [spawn](https://api.dart.dev/stable/2.7.1/dart-isolate/Isolate/spawn.html) a new isolate / event loop? And call into our native library?

vaind: Haven't tried that yet and I think it's just like threads in any other language (except you can't share variables inside dart). 

And yes tests like that make sense. We had an issue regarding isolates so I kinda related stuff like this to that issue, in my head. writing it down now #52

greenrobot: If isolates behave like threads and vars like thread-locals, this would be a bug, agreed?

vaind: how so? I don't see a problem

greenrobot: To my limited understanding, I imagine this to happen:
- Let's assume two isolates, both starting with IDs at 1 (produce colliding IDs, because isolated...)
- Callbacks from core happen on a different thread
- (Open question: not sure about how a "foreign" thread is mapped to an isolate. So, which isolate gets the callback?)
- Now, I don't see how we ensured that one isolate gets **only its callback in using its ID**.

E.g. Isolate 2 registers a callback for ID 1, but isolate 1 gets to consume it because it also has a pending callback for ID 1.

vaind: > Callbacks from core happen on a different thread

They don't with objectbox-c - all calls are synchronous, executed on the same thread. This approach wouldn't work for other kinds of callbacks, where you register the callback and it may happen in the future (e.g. some kind of event listeners). Not sure how that could be handled with dart, if at all.

vaind: Looks like it's an open issue with some workarounds available https://github.com/dart-lang/sdk/issues/37022

greenrobot: Right, that one is strictly synchronous and if FFI guarantees synchronous callbacks to execute in the same isolate (I assume that is defined somewhere?), we should be fine.

PR #108, opened 2020-06-23
Add support for property queries.

greenrobot-team:
Squashed and rebased changes from #75 

## Remaining issues
- See #75 for remaining issues.
  - ~`.find(replaceNullWith:-1)` broken for negative values: it looks like unsigned integers are returned? E.g. `255/65535/4294967295/-1` is returned instead of `-1` for `8/16/32/64` bit integers.~ Wrong types in `Struct`s.
  - Byte unsigned issue is rooted in the binding treating them as unsigned, submitted patch in #109. 
  - ~Similar: floats return `0.0` instead of replacement value, but double works fine.~ Was casting double to float.
- ~There are code errors in bindings.dart and signatures.dart.~
- ~Native property query appears to auto-close/free, e.g. getting results twice crashes, or closing/freeing the property query crashes.~ Gone, probably wrong pointer types? Odd.
- Missing from tests: edge cases (e.g. min/max integer, NaN float, etc.).


lib/src/bindings/structs.dart
```diff
@@ -118,3 +118,76 @@ class OBX_bytes_array extends Struct {
...
+
+class OBX_double_array extends Struct {
+
+  Pointer<Double> _itemsPtr;
+
+  @IntPtr() // size_t
+  int count;
+
+  List<double> items() => _itemsPtr.asTypedList(count).toList();
+}
```
vaind, line 193: These could probably be replaced with a single template (generic)

lib/src/query/property.dart
```diff
@@ -0,0 +1,227 @@
...
+    }
+  }
+}
+
+/// shared implementation, hence mixin
+mixin _CommonNumeric on PropertyQuery {
+
+  double average() {
+    final ptr = allocate<Double>();
+    checkObx(bindings.obx_query_prop_avg(_cProp, ptr));
```
vaind, line 51: AFAICT this should be inside try-catch because `ptr` is already allocated and must be freed even if checkObx throws

vaind: also check other instances of this pattern

lib/src/query/property.dart
```diff
@@ -0,0 +1,227 @@
...
+  List<int> _unpack64(Pointer<OBX_int64_array> ptr) {
+    try {
+      return ptr.ref.items();
+    }finally {
+      bindings.obx_int64_array_free(ptr);
+    }
+  }
+
+  List<int> find({int replaceNullWith}) {
+    final ptr = replaceNullWith != null ? (allocate<Int64>()..value = replaceNullWith) : Pointer<Int64>.fromAddress(0);
```
vaind, line 118: looks like another memory leak

lib/src/query/property.dart
greenrobot-team: Done. Though still `Attempt to free invalid pointer 0x8` presumably for `test/query_property_test.dart: .sum integers`. Knowing the exact line would be nice...

lib/src/query/property.dart
greenrobot-team: The pointer is freed within `_curryWithDefault` protected against `checkObxPtr` throwing, so fine?

lib/src/query/property.dart
greenrobot-team: OK, not closing the property query pointer helps. Investigating.

lib/src/bindings/structs.dart
greenrobot-team: Tried this, but Struct can't be extended by a generic class. Using a mixin then. But then `.asTypedList(count)` is not available. Anything else in mind?

greenrobot-team:
@vaind Some other issues: it looks like for 8-32 bit integers, if their value is negative the property query returns an unsigned value. When doing sums or average they also appear to be treated as unsigned.

Also byte values appear to get returned as unsigned, even via the box interface. Though the sum then works correctly. Oh my.

greenrobot-team:
Resolved integer issue by correcting `Struct`s of int array types to use signed integer pointers.

greenrobot-team:
Byte unsigned issue is rooted in the binding treating them as unsigned, submitted patch in #109. 

This leaves one issue: null floats (not doubles) getting returned as `0.0`.

greenrobot-team:
Was casting double to float, issue resolved as well.

Pending #109 and a final round of clean-up and docs (e.g. on find(), sum() and etc. methods).

lib/src/query/property.dart
greenrobot-team: Updating here, too: closing seems to work now, for whatever reason.

test/index.dart
```diff
@@ -0,0 +1,9 @@
+import 'box_test.dart' as box_test;
+import 'query_test.dart' as query_test;
+import 'query_property_test.dart' as query_property_test;
+
+void main() {
+  box_test.main();
+  query_test.main();
+  query_property_test.main();
+}
```
greenrobot-team, line 9: @vaind Can this file be removed? Not sure why this was added by @Buggaboo.

vaind: Yeah, definitely. It wasn't added lately - it's a leftover from the old state of main/dev (you can see that in the file's history)

vaind:
@greenrobot-team see the latest commits - should be ready for the merge after #109 is merged so that the tests pass

vaind:
Just for future reference - the issue with double-free/invalid pointers was caused by a signature mismatch between dart property query function definitions and c-api functions, e.g. `obx_query_prop_sum` - it was missing the last argument...

greenrobot-team:
Rebased onto main with #109 included. Tests all good now. Merging.

PR #109, opened 2020-06-29
Unmarshal byte as Int8 (was Uint8).

greenrobot-team:
To be consistent with property queries and Java unmarshal bytes as signed integers.

Prerequesite for #108.

@vaind Can you have a look? I doubt this change is without side-effects. Or maybe totally on the wrong track.


greenrobot-team:
Merging this then.

PR #120, opened 2020-09-26
WiP: @Unique and @Index (Any help is welcome)

Buggaboo:
I have a basic implementation, but the tests keep exploding with this though:

```
ObjectBoxException: failed to create store: 10501 No index ID/UID assigned for indexed property: Property uByte (15, Byte)
```

This error applies to all the `@Unique` annotated fields.

Afaik objectbox-model.json seems normal.

Help!


greenrobot-team:
Just some quick feedback: `@Unique` is actually just a special `@Index`. So this AFAIK won't work without adding support for indexes first (at least in other supported languages). An index needs to get assigned an id/uid and be stored in the model file.

greenrobot-team:
None of our examples use an index, yet. I should change that. ^^

OK, more details: the model file needs an `indexId` for each property. The last index id ever used should also be set as `lastIndexId`. Retired index UIDs need to move to `retiredIndexUids`.
```json
...
        {
          "id": "13:8914743272487704330",
          "name": "dateProp",
          "indexId": "12:1548007883894887437",
          "type": 10,
          "flags": 8
        }
...
  "lastIndexId": "12:1548007883894887437",
...
  "retiredIndexUids": [],
```

So this is actually a lot more work than just adding the `UNIQUE` or `INDEXED` flag.

Buggaboo:
What's the difference between id and uid?

greenrobot-team:
> What's the difference between id and uid?

IDs are given out in order (`1, 2, 3...`). The UID is a random number that identifies something over the lifetime of the Store.
https://docs.objectbox.io/advanced/meta-model-ids-and-uids goes into more details on this.

Side note: I updated the Kotlin example to use an index.
https://github.com/objectbox/objectbox-examples/tree/main/android-app-kotlin

Buggaboo:
It was a lot of work... there should be a unified way to setup the models.

```
ObjectBoxException: failed to create store: 10001 State condition failed in assignIdsForProperty:476: indexId <= catalog.lastIndexId()
```

I have no idea what the above means, but its breaking everything.
I did stupid stuff like indexing booleans etc. only I got an OBException in runtime about the indexed float types.

Help!

lib/src/bindings/bindings.dart
```diff
@@ -292,6 +299,9 @@ class _ObjectBoxBindings {
     obx_model_last_entity_id =
         _fn<obx_model_last_entity_id_native_t>('obx_model_last_entity_id')
             .asFunction();
+    obx_model_last_index_id =
+        _fn<obx_model_last_index_id_native_t>('obx_model_last_entity_id')
```
greenrobot-team, line 303: > 'obx_model_last_entity_id'

That's wrong? This probably explains the error thrown.

greenrobot-team:
Thanks!

It's odd that `objectbox.h` does not document types that do not support an index. Peeking at the Java docs: https://docs.objectbox.io/entity-annotations#property-indexes-with-index Java byte arrays, float and double can't be indexed.

test/box_test.dart
```diff
@@ -52,6 +52,15 @@ void main() {
     expect(item.tString, equals('Two'));
   });
 
+  test('.put() cannot add duplicate values on a unique field', () {
+    final u1    = TestEntity.unique(uString: 'a', uLong: 1, uInt: 1,
+        uShort: 1, uBool: false, uByte: 1, uChar: 1);
+    final again = TestEntity.unique(uString: 'a', uLong: 1, uInt: 1,
+        uShort: 1, uBool: false, uByte: 1, uChar: 1);
+
+    expect(box.putMany([u1, again]), throwsException);
```
greenrobot-team, line 61: Should check the exception message, no?

test/entity.dart
```diff
@@ -63,4 +64,68 @@ class TestEntity {
     omit = -1;
     disregard = 1;
   }
+
+  @Unique(type:OBXPropertyType.Byte)
```
greenrobot-team, line 68: `@Unique` and `@Index` should be annotations in addition to `@Property`, no?

greenrobot-team: Oh, wait. I thought in Dart `@Property` is always required, but no. Still, if defining a type or UID it should be done with `@Property`, no? And there should be no need to expose `indexUid` to the user.

generator/lib/src/entity_resolver.dart
```diff
@@ -87,6 +90,18 @@ class EntityResolver extends Builder {
...
+        propUid = _uniqueAnnotation.getField('uid').toIntValue();
+        indexUid = _uniqueAnnotation.getField('indexUid').toIntValue();
+        fieldType = _uniqueAnnotation.getField('type').toIntValue();
+        flags = OBXPropertyFlag.UNIQUE;
+      } else if (_indexChecker.hasAnnotationOfExact(f)) {
+        final _indexAnnotation = _indexChecker.firstAnnotationOfExact(f);
+        propUid = _indexAnnotation.getField('uid').toIntValue();
+        indexUid = _indexAnnotation.getField('indexUid').toIntValue();
+        fieldType = _indexAnnotation.getField('type').toIntValue();
+        flags = OBXPropertyFlag.INDEXED;
```
greenrobot-team, line 104: Hm, I see why it would make sense to have `@Unique` and `@Index` be exclusive annotations. However, they may be used both on one property. Also, as said on `test/entity.dart` they should not replace `@Property`.

Compare with Java: https://docs.objectbox.io/entity-annotations#property-indexes-with-index ff
The index type may be further configured, even if using `@Unique`. But we/I can also look into implementing this.

greenrobot-team:
Thanks for getting this started! If time permits it I might continue on this tomorrow. It's kind of hard to get this right without any docs on how this should work internally.

lib/src/bindings/bindings.dart
Buggaboo: Thanks, I tend to do that too often, tired eyes.

test/box_test.dart
Buggaboo: That was more or less a mindfart placeholder. We can write proper tests now.

test/entity.dart
Buggaboo: Could indexUid be part of a back end front end contract?

generator/lib/src/entity_resolver.dart
Buggaboo: I'm out of my depth here...

test/entity.dart
vaind: I'd keep it the same way [Java does it, i.e. separate annotations](https://docs.objectbox.io/entity-annotations#property-indexes-with-index). The code's already updated in such a way so I'm resolving this conversation.

generator/lib/src/entity_resolver.dart
vaind: outdated conversation. done like in Java (separate annotations, index type), resolving the conversation.

test/box_test.dart
vaind: I've fixed the test

vaind:
replaced by #123

PR #123, opened 2020-10-06
Support indexes, add Index and Unique annotation

greenrobot-team:
Continuation of #120 

Making progress.

Overall I'm not sure if there should be separate annotations vs. using flags on the existing `@Property` annotation.


Buggaboo:
Unrequested review: I think my `isIndexer` extension, needs updating to include the other types.

greenrobot-team:
@Buggaboo Yeah, thanks. This is not done, yet. Hence the draft status.

Buggaboo:
My commit b5939d8a8 (aka "Why merge twice?") was a bad idea, in hindsight.

That needs to be reverted.

That function call is doing more than merging, it also puts unused uids in their respective 'retired' lists. Imho, the function is too fat, it should be split up.

The proper place to set the `lastIndexProperty` should be:
```dart
IdUid mergeEntity(ModelInfo modelInfo, ModelEntity entity) {

	// ...

    final indexProps =
        entityInModel.properties.where((p) => p.type.isIndexer).toList();

    if (indexProps.isNotEmpty) {
      modelInfo.lastIndexId = indexProps.last.id;
    }

    return entityInModel.id;
  }
```

Also, `retiredIndexUid` should be looked at.

vaind:
Fixed issues, added tests and cleaned up. fast-tracking this to get on with other PRs but please have a look at the changes @greenrobot-team

greenrobot-team:
@vaind Invalid usages of the annotations and index type are not tested, otherwise LGTM. Thanks for finishing this! (You even caught all TODOs and left-over code removals!)

Issue #125, opened 2020-10-12
Support query with no condition


greenrobot-team:
Currently building a query requires specifying at least one condition. E.g. to use the upcoming query observer with an all objects query, this results in strange code like

```dart
final dummyCondition = dateProp.greaterThan(0);
_query = _box.query(dummyCondition)
```

Support building a query with no conditions (AFAIK at least OBX for Java allows this as well).

Buggaboo:
Maybe something like:

```dart

// ctor
Box.emptyQuery() { /* return an empty query, skip the queryBuilder */ }
```

PR #132, opened 2020-10-22
Support for conditionless queries

Buggaboo:
This should help solve #125.

Please review.

### DIFF
--- a/lib/src/box.dart
+++ b/lib/src/box.dart
@@ -257,6 +257,10 @@ class Box<T> {
     }
   }
 
+  /// When no conditions are required on properties
+  QueryBuilder get emptyQuery =>
+      QueryBuilder<T>(_store, _fbManager, _modelEntity.id.id);
+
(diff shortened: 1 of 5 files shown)


lib/src/box.dart
```diff
@@ -257,6 +257,10 @@ class Box<T> {
     }
   }
 
+  /// When no conditions are required on properties
+  QueryBuilder get emptyQuery =>
+      QueryBuilder<T>(_store, _fbManager, _modelEntity.id.id);
+
```
vaind, line 263: Not sure we need a new method/getter. In some other language bindings it's just box.query() with no arguments.

vaind (changes requested):

lib/src/box.dart
Buggaboo: Square brackets would do the job.

```dart
QueryBuilder get query(..., [Condition c]) =>
```

Feel free to make the modification.

PR #133, opened 2020-10-22
WiP, Preview: relations

Buggaboo:
@greenrobot-team I need some help setting up many-to-many relations.

### DIFF
--- a/generator/integration-tests/common.dart
+++ b/generator/integration-tests/common.dart
@@ -59,13 +59,18 @@ commonModelTests(ModelDefinition defs, ModelInfo jsonModel) {
         jsonModel.retiredEntityUids);
   });
 
-  // TODO when indexes are available
+// TODO when indexes are available
+// TODO see one-liner in the next lastRelationId test
 //  test('lastIndexId', () {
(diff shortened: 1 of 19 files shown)


vaind (commented):
I've had a look at your proposal and while some ideas look good, I've also have found a couple of issues:
* model manipulation, such as managing UIDs, relations, indexes, should happen in `lib/modelinfo/model*.dart`, not directly in code_builder
* the C-binding used is based on the (old) way of defining functions - manually, while the repo uses generated code already,
* some string pointer (memory) management issues

Since this is based on an old commit, I think I'll restart from the current `main` and cherry-pick what makes sense. Hopefully the commits are structured to allow reasonable cherry-picking

vaind:
And I've noticed relation properties (to-one rels) & standalone relations (to-many) were kinda mixed up here - lastRelationId and retiredRelationIds are only relevant for standalone relations while this PR only starts to implement relation properties.

I've cherry-picked and squashed the relevant parts here: 09cce1ba2d2cb2d6a867b1c323482d8d61b6e0fd and will continue on that branch.

Issue #142, opened 2020-11-06
Observers broken when multiple isolates are used


vaind:
Current observer implementation (used by query streams) is broken if the data is created by more than one isolate. In that case, the observer callback from the native side is executed in a "wrong" isolate, crashing the program.

Test-case to reproduce: https://github.com/objectbox/objectbox-dart/blob/async-and-isolates/test/isolates_test.dart#L85 
`error: Cannot invoke callback on incorrect isolate.`

This is really an issue affecting any **async** callbacks from the native side, I'll create a followup to address it in a generic way.

In case the Store is only used from a single isolate, it should work fine AFAICT because the callback is issued from the same isolate. However, with the introduction of SyncClient, the observer may be called asynchronously even though lib user only uses a single isolate so this must be addressed.

vaind:
On the same not, but probably unrelated: resource management seems a little hard to do with Query.findStream() - the query itself must be closed at some point to avoid memory leak. e.g. `stream_test.dart` don't follow this...

vaind:
Works fine (at least the isolates test that previously failed does) in #145

vaind:
Resolved by #145

PR #145, opened 2020-11-07
Async callbacks & multi-isolate store access

vaind:
The changes necessary are pretty simple - just allow creating Store from an existing `OBX_Store` C pointer and pass that between isolates. Any ideas how to improve/simplify that for users? I don't really know the most common isolates' usage patterns in Dart/Flutter yet so I'm open to suggestions.

However, to get it working, we had to do some custom stuff to send handle async callbacks (issue  #142) - objectbox-c now (v0.12+) includes dart-specific functionality to connect to a dart runtime and send callback data over streams. This required a rework of the observers.

Besides the above, this PR also implements sync listeners.

### DIFF
--- a/.github/workflows/dart.yml
+++ b/.github/workflows/dart.yml
@@ -26,10 +26,8 @@ jobs:
           - macos-10.15
           - ubuntu-20.04
         dart:
-          - 2.10.5
-          # - 2.9.3 - generator stuck. I remember there was an issue in some dependency but don't remember which one.
-          - 2.8.4
-          - 2.7.2
+          - latest
+          - 2.10.0 # currently the lowest fully supported version (i.e. generator + lib)
     runs-on: ${{ matrix.os }}
     steps:
       # Note: dart-sdk from flutter doesn't work on linux, see https://github.com/flutter/flutter/issues/74599
--- a/benchmark/pubspec.yaml
+++ b/benchmark/pubspec.yaml
@@ -2,7 +2,7 @@ name: objectbox_benchmark
 description: Simple ObjectBox-Dart performance benchmark
 
 environment:
-  sdk: ">=2.6.0 <3.0.0"
+  sdk: ">=2.10.0 <3.0.0"
 
 dependencies:
   objectbox: any
--- a/flutter_libs/android/build.gradle
+++ b/flutter_libs/android/build.gradle
@@ -12,5 +12,5 @@ android {
 
 dependencies {
     // https://bintray.com/objectbox/objectbox/io.objectbox%3Aobjectbox-android
-    implementation "io.objectbox:objectbox-android:2.8.0"
+    implementation "io.objectbox:objectbox-android:2.9.0"
 }
--- a/flutter_libs/ios/download-framework.sh
+++ b/flutter_libs/ios/download-framework.sh
@@ -4,19 +4,21 @@ set -euo pipefail
 # NOTE: run this script before publishing
 
 # https://github.com/objectbox/objectbox-swift/releases/
-obxSwiftVersion="1.4.1"
+obxSwiftVersion="1.5.0-beta1"
 
 dir=$(dirname "$0")
 
-url="https://github.com/objectbox/objectbox-swift/releases/download/v${obxSwiftVersion}/ObjectBox-framework-${obxSwiftVersion}.zip"
+url="https://github.com/objectbox/objectbox-swift-spec-staging/releases/download/v1.x/ObjectBox-xcframework-${obxSwiftVersion}.zip"
 zip="${dir}/fw.zip"
 
 curl --location --fail --output "${zip}" "${url}"
 
+frameworkPath=Carthage/Build/ObjectBox.xcframework/ios-arm64/ObjectBox.framework
+
 rm -rf "${dir}/Carthage"
 unzip "${zip}" -d "${dir}" \
-  "Carthage/Build/iOS/ObjectBox.framework/Headers/*" \
-  "Carthage/Build/iOS/ObjectBox.framework/ObjectBox" \
-  "Carthage/Build/iOS/ObjectBox.framework/Info.plist"
+  "${frameworkPath}/Headers/*" \
+  "${frameworkPath}/ObjectBox" \
+  "${frameworkPath}/Info.plist"
 
 rm "${zip}"
--- a/flutter_libs/pubspec.yaml
+++ b/flutter_libs/pubspec.yaml
@@ -5,8 +5,8 @@ homepage: https://objectbox.io
 description: ObjectBox is a super-fast NoSQL ACID compliant object database. This package contains flutter runtime libraries for ObjectBox.
 
 environment:
-  sdk: ">=2.6.0 <3.0.0"
-  flutter: ">=1.12.0 <2.0.0"
+  sdk: ">=2.9.0 <3.0.0"
+  flutter: ">=1.20.0 <2.0.0"
 
 dependencies:
   # This is here just to ensure compatibility between objectbox-dart code and the libraries used
--- a/generator/integration-tests/shared-pubspec.yaml
+++ b/generator/integration-tests/shared-pubspec.yaml
@@ -1,7 +1,7 @@
 name: objectbox_generator_test
 
 environment:
-  sdk: ">=2.5.0 <3.0.0"
+  sdk: ">=2.10.0 <3.0.0"
 
 dependencies:
   objectbox: any
--- a/generator/pubspec.yaml
+++ b/generator/pubspec.yaml
@@ -5,7 +5,8 @@ homepage: https://objectbox.io
 description: ObjectBox binding code generator - finds annotated entities and adds them to the ObjectBox DB model.
 
 environment:
-  sdk: ">=2.5.0 <3.0.0"
+  # min SDK v2.10.0 (or Flutter v1.22) - there were breaking changes in the analyzer/resolver
+  sdk: ">=2.10.0 <3.0.0"
 
 dependencies:
   objectbox: 0.11.0
--- a/install.sh
+++ b/install.sh
@@ -8,7 +8,7 @@ set -eu
 #   * update lib/src/bindings/objectbox.h
 #   * execute pub run ffigen
 #   * have a look at the changed files to see if some call sites need to be updated
-cLibVersion=0.11.0
+cLibVersion=0.12.0
 os=$(uname)
 
 # if there's no tty this is probably part of a docker build - therefore we install the c-api explicitly
--- a/objectbox/CHANGELOG.md
+++ b/objectbox/CHANGELOG.md
@@ -1,3 +1,9 @@
+## latest
+
+* Update to objectbox-c v0.12.0
+* Update to objectbox-android v2.9.0
+* Update to objectbox-swift v1.5.0
+
 ## 0.11.0 (2021-02-01)
 
 * Add `ToOne<>` class to wrap related entities. See examples for details.
--- a/objectbox/CHANGELOG.md
+++ b/objectbox/CHANGELOG.md
@@ -6,6 +6,7 @@
 * Update to objectbox-c v0.12.0
 * Update to objectbox-android v2.9.0
 * Update to objectbox-swift v1.5.0
+* Increase minimum SDK versions: Flutter v1.20 & Dart v2.9. Code generator already required Flutter v1.22 & Dart v2.10. 
 
 ## 0.11.0 (2021-02-01)
 
--- a/objectbox/CHANGELOG.md
+++ b/objectbox/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## latest
 
+* Add `Store.reference` getter and `Store.fromReference()` factory - enabling access to store from multiple isolates.
+* Add `Store.subscribe<EntityType>()` and `Store.subscribe<EntityType>()` data change event streams.
+* Add multiple `SyncClient` event streams.
 * Update to objectbox-c v0.12.0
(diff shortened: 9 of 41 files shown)


Buggaboo:
Good solution!

lib/src/observable.dart
```diff
@@ -1,92 +1,131 @@
...
-      unsubscribe(store);
-    });
+  _Observer() {
+    initializeDartAPI();
+  }
+
+  // start() is called whenever user starts listen()-ing to the stream
+  void init(void Function() start) {
+    controller = StreamController<StreamValueType>(
+        onListen: start, onPause: stop, onResume: start, onCancel: stop);
```
RTrackerDev, line 34: `onListen` works for each new listener, and `onCancel` when all unsubscribed. is this taken into account? It would be nice if you wrote a test of how many times a method with multiple subscribers is called and whether all `_cObserver` are closed before being reinstalled.

lib/src/observable.dart
```diff
@@ -1,92 +1,131 @@
...
+    final observer = _Observer<void>();
+    final entityId = entityDef<EntityT>().model.id.id;
+
+    observer.init(() {
+      // We're listening to events on single entity so there's no argument.
+      // Ideally, controller.add() would work but it doesn't, even though we're
+      // using StreamController<Void> so the argument type is `void`.
+      observer.receivePort = ReceivePort()
+        ..listen((_) => observer.controller.add(null));
+      observer.cObserver = bindings.obx_dart_observe_single_type(
```
RTrackerDev, line 67: I think you should close previous if `cObserver != null`. this is if you want to subscribe multiple times to the same thread.In this case, `onListen` will be called several times and` onCancel` only when the last one is unsubscribed.

lib/src/observable.dart
```diff
@@ -1,92 +1,131 @@
...
+          });
+        });
+      observer.cObserver = bindings.obx_dart_observe(ptr, observer.nativePort);
+    });
+
+    return observer.stream;
+  }
+}
+
+extension Streamable<T> on Query<T> {
```
RTrackerDev, line 122: Could you add method `countStream`?

lib/src/observable.dart
```diff
@@ -1,92 +1,131 @@
...
+/// StreamController implementation inspired by the sample controller sample at:
+/// https://dart.dev/articles/libraries/creating-streams#honoring-the-pause-state
+/// https://dart.dev/articles/libraries/code/stream_controller.dart
+extension ObservableStore on Store {
+  /// Create a stream to data changes on EntityT (stored Entity class).
+  ///
+  /// The stream receives an event whenever an object of EntityT is created or
+  /// changed or deleted. Make sure to close() the subscription after you're
+  /// done with it to avoid hanging change listeners.
+  Stream<void> subscribe<EntityT>() {
```
RTrackerDev, line 57: I believe you should store `observers` for entity type locally and reuse them.

lib/src/observable.dart
Buggaboo: You could also count on the query from the queryStream, right?

RTrackerDev: For example:
```
Stream<int> get countStream => store.subscribe<T>().map((_) => count());
```
@Buggaboo As I understand `count()`, this is a lighter operation than pull all the data with `find`

RTrackerDev: It would also be great to have:
```
Stream<T> get findFirstStream => store.subscribe<T>().map((_) => findFirst());
```

lib/src/observable.dart
vaind: AFAIK you can only listen to a stream once (it's not a broadcast stream)

RTrackerDev: Then there is another problem with accumulation `StreamController` for every `query` stream in memory

lib/src/observable.dart
vaind: yes, planning to take a look at that

lib/src/observable.dart
vaind: same as above: https://github.com/objectbox/objectbox-dart/pull/145#discussion_r535180104

lib/src/observable.dart
```diff
@@ -1,92 +1,131 @@
...
+    final observer = _Observer<void>();
```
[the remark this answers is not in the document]
vaind: same as here: https://github.com/objectbox/objectbox-dart/pull/145#discussion_r535184003

lib/src/observable.dart
RTrackerDev: > AFAIK you can only listen to a stream once (it's not a broadcast stream), I'll check though.

@vaind What if user do: `query.findStream().asBroadcastStream(onListen: (s){}, onCancel: (s)=> s.cancel())` in this case this methods will be called in the prime `StreamController` too as i remember. I think you should check it.

lib/src/observable.dart
RTrackerDev: This is just a suggestion for improvement.

Buggaboo: Of course. I appreciate it. You could branch off from @vaind, and offer a PR to his branch, I suspect he's too busy atm.

RTrackerDev: @Buggaboo So I understand. And if @vaind doesn't mind I can create PR for this branch.

vaind: would wait with that, I'm still changing & more importantly rebasing this so it would mess up your fork

test/isolates_test.dart
```diff
@@ -0,0 +1,147 @@
...
+
+    // Send a message to the isolate
+    expect(await call('hello'), equals('re:hello'));
+    expect(await call('foo'), equals('re:foo'));
+
+    isolate.kill(priority: Isolate.immediate);
+    receivePort.close();
+  });
+
+  /// Work with a single store accross multiple isolates
```
greenrobot-team, line 49:
```suggestion
  /// Work with a single store across multiple isolates
```

Proof that I read this file. 🤓 

(jk, IDE highlights spelling errors)

test/observer_test.dart
```diff
@@ -88,94 +23,128 @@ void main() async {
...
-    final callback =
-        Pointer.fromFunction<obx_observer_single_type>(callbackSingleType);
-    final observer = bindings.obx_observe_single_type(
-        store.ptr, testEntityId, callback, randomPtr);
+      if (expectedEvents == 0) {
+        completer.complete();
+      }
+    });
 
+    // expect two events after one put() and one putMany()
```
greenrobot-team, line 90: Maybe here, and in above test as well change the note so sth. like "expect to receive events for both entity types" (and above e.g. "expect to receive events for only TestEntity").

Anyhow, should be "three events".

lib/src/observable.dart
```diff
@@ -1,85 +1,138 @@
...
-    StoreCloseObserver.removeListener(store, _anyObserver[storeAddress]);
-    bindings.obx_observer_close(_anyObserver[storeAddress]);
-    _anyObserver.remove(storeAddress);
-    syncOrObserversExclusive.unmark(store);
   }
 
-  static bool isSubscribed(Store store) =>
-      _Observable._anyObserver.containsKey(store.ptr.address);
+  void _debugLog(String message) {
+    // print('Observer=${_cObserver?.address} $message');
```
greenrobot-team, line 53: Is this a left-over from testing? This is still a draft so maybe don't care...

test/observer_test.dart
```diff
@@ -88,94 +23,128 @@ void main() async {
...
+      await completer.future.timeout(defaultTimeout);
+
+      // won't trigger when paused
+      subscription.pause();
+      completer = Completer();
+      box.put(simpleStringItems[0]);
+      expect(completer.future.timeout(defaultTimeout),
+          throwsA(isA<TimeoutException>()));
+
+      // triggers when resumed
```
greenrobot-team, line 133: Maybe note that it only triggers once, not twice (no buffered event as the underlying Observable should stop events while stream is paused).

lib/src/observable.dart
```diff
@@ -1,85 +1,138 @@
...
 
-    _Observable._any[storeAddress] ??= <int, Any>{};
-    _Observable._any[storeAddress] /*!*/ [entityId] ??= (u, _, __) {
-      // dummy value to trigger an event
-      _Observable.controller.add(entityId);
-    };
+    observer.init(() {
+      // We're listening to events on single entity so there's no argument.
+      // Ideally, controller.add() would work but it doesn't, even though we're
+      // using StreamController<Void> so the argument type is `void`.
```
greenrobot-team, line 73: Maybe add an `EmptyEvent` then?

But yeah. Does not seem to be a popular request though: https://github.com/dart-lang/sdk/issues/34176

lib/src/observable.dart
```diff
@@ -1,85 +1,138 @@
...
+      observer.cObserver = bindings.obx_dart_observe_single_type(
+          ptr, entityId, observer.nativePort);
+    });
+
+    return observer.stream;
+  }
+
+  /// Create a stream to data changes on all Entity types.
+  ///
+  /// The stream receives an even whenever any data changes in the database.
```
greenrobot-team, line 85:
```suggestion
  /// The stream receives an event whenever any data changes in the database.
```

lib/src/sync.dart
```diff
@@ -211,6 +232,271 @@ class SyncClient {
...
+    assert(!finished, 'finish() may only be called once.');
+    controller = StreamController<StreamValueType>.broadcast(
+        onListen: _start,
+        /* not for broadcast streams: onPause: _stop, onResume: _start,*/
+        onCancel: _stop);
+    finished = true;
+    return controller.stream;
+  }
+
+  // stop() is called when the stream subscription is started or resumed
```
greenrobot-team, line 441:
```suggestion
  // start() is called when the stream subscription is started or resumed
```

lib/src/sync.dart
```diff
@@ -211,6 +232,271 @@ class SyncClient {
...
+      _loginEvents.add(_SyncListenerConfig(
+          (int nativePort) =>
+              bindings.obx_dart_sync_listener_login(ptr, nativePort),
+          (_, controller) => controller.add(SyncLoginEvent.loggedIn)));
+
+      _loginEvents.add(_SyncListenerConfig(
+          (int nativePort) =>
+              bindings.obx_dart_sync_listener_login_failure(ptr, nativePort),
+          (code, controller) {
+        // see OBXSyncCode - TODO should we match any other codes?
```
greenrobot-team, line 281: FYI Java just passes the code here, so 🤷

test/sync_test.dart
```diff
@@ -175,29 +159,177 @@ void main() {
...
+        expect(read1 /*!*/ .id, equals(read2 /*!*/ .id));
+        expect(read1 /*!*/ .tLong, equals(read2 /*!*/ .tLong));
+      });
+
+      test('SyncClient listeners: connection', () async {
+        final client = createClient(env.store);
+
+        // collect connection events
+        final events = <SyncConnectionEvent>[];
+        final streamSub = client.connectionEvents.listen(events.add);
```
greenrobot-team, line 194: Is `streamSub` not cancelled at the end of the test on purpose?

test/sync_test.dart
```diff
@@ -175,29 +159,177 @@ void main() {
...
-      TestEntity /*?*/ read2 = env2.box.get(id);
-      expect(read1, isNotNull);
-      expect(read2, isNotNull);
-      expect(read1 /*!*/ .id, equals(read2 /*!*/ .id));
-      expect(read1 /*!*/ .tLong, equals(read2 /*!*/ .tLong));
+        final client2 = loggedInClient(env2.store);
+        await client2.completionEvents.first.timeout(Duration(seconds: 1));
+        client2.close();
+
+        expect(env2.box.get(1) /*!*/ .tLong, 100);
```
greenrobot-team, line 278: Should this be:
```suggestion
        expect(env2.box.get(id) /*!*/ .tLong, 100);
```

greenrobot-team:
LGTM as far as I understand (no deep knowledge of streams and isolates; only had a glimpse at C API for Dart).

lib/src/observable.dart
vaind: Nope, the following code calls underlying stream onCancel() only once. 
```dart
final queryStream = query.findStream().asBroadcastStream(onListen: (s) { }, onCancel: (s) => s.cancel(); });
final subscription = queryStream.listen((list) {});
final subscription2 = queryStream.listen((list) {});
await subscription.cancel();
await subscription2.cancel();
```
Please create an issue (ideally with a reproducible example) if you find out observers/query streams don't work as expected.

lib/src/observable.dart
vaind: Recently, I've been looking at this a little differently, why create a stream-specific method for everything when you can do the same easily with query.stream.map(), I've created a follow up issue #195

RTrackerDev: > Recently, I've been looking at this a little differently, why create a stream-specific method for everything when you can do the same easily with query.stream.map(), I've created a follow up issue #195

But query stream always fetch data...

greenrobot: @RTrackerDev thanks for your comment, let's continue on #195

lib/src/observable.dart
vaind: Follow up issue #195

lib/src/observable.dart
vaind: On one hand, yes, on the other, it's more like a future-proofing the code for when you need to check something manually while developing :)

lib/src/observable.dart
vaind: IDK, looks even worse to me

Issue #81, opened 2019-12-02
Mobile app integration test CI


vaind:
We should have a way to test ObjectBox runs on mobile (simulator) Android and iOS. From the quick look in the past I think our current tests (/test folder) can't be just executed on the target platform and we would rather need to have an app that instead runs the test suite itself. I've done this quickly in the past with the `/lib/integration_test.dart`. If this is the right approach, it should be moved to a separate dart/flutter project and run from CI for both mobile platforms (when iOS support #47 is merged)

vaind:
Ideally, we should run the complete test suite (`test` folder) on all supported devices, if possible.

vaind:
I've partially prepared this: there's a `flutter-integration-test` job running in GH Actions. Right now, it builds a release apps of both demo applications (sync & non-sync) on mac, windows and Linux.

TODOs:
* run `flutter drive`: Flutter 2.0 has come out with a non-null-safe version of flutter_driver, basically turning it unusable, despite being built-in the SDK :roll_eyes:. As soon as a null-safe version of flutter_driver comes out (hopefully with Flutter 2.1?) we can have a look at the CI and see if it's an option to launch emulator/simulator. If so, `flutter drive` should just work and we can reenable it in `./objectbox/tool/integration-test.sh`
* still untouched - try running the whole test-suit in an app on device. I suspect this could be run with just `lib/main.dart` - looks like a Flutter app doesn't actually have to have any flutter code to be able to run? Let's see...

roiskhoiron:
Is this task already closed? @vaind

greenrobot-team:
@roiskhoiron No, this is still open. Also note that vaind no longer works on this project, so avoid tagging!

RaghvindYadav:
@greenrobot-team Hey I can help your team to resolve this issue. I worked a lot in flutter integrating, widget and unit tests. I also did widget-book implementation.

greenrobot-team:
@RaghvindYadav Thanks for offering! We would welcome a contribution (make sure . As described in the requirements above, do you know if and how the existing unit tests could be run? Is it even possible to test on Android emulators and iOS simulators on GitHub Actions? Would a small proof of concept be a good starting point?

Edit: and maybe have a look at [the contributing info](https://github.com/objectbox/objectbox-dart/blob/main/CONTRIBUTING.md) and make sure you agree that your code would use [the license of this project](https://github.com/objectbox/objectbox-dart/blob/main/LICENSE).

[...]

PR #146, opened 2020-11-10
Use ffigen for C-API bindings

### LINKED ISSUE #139: Clean up signatures.dart & bindings.dart
`signatures.dart` contains C-API signatures and those are used in DynamicLibrary lookup in `bindings.dart`. Those files have became a little large and repetitive. Let's think about how they can be cleaned up. Some ideas:

* use typedefs for common signatures for different "classes", like store, box, ...
* maybe rename "bindings" variable to something shorter, e.g. I like Go's `C.`
* maybe drop the `obx_` prefix

vaind:
Part of #139 - switches from manual binding definitions to `ffigen` generated code


vaind:
@greenrobot-team you can have a first look when it works for you. Tests crash at the moment but I don't expect many changes to get it fixed.

greenrobot-team:
Won't be able to complete my review until I'm back. Anyhow, most of what I've seen of the first commit seems straight forward.

Buggaboo:
You might want to check the size of the pointers. They should be machine word size. Struct padding (I assume no packing is applied) could also be an issue.

Really cool tool. I didn't know about it until now. I tried writing my own this summer. I guess I could throw that away now.

vaind:
> You might want to check the size of the pointers. They should be machine word size, unless instructed differently to the compiler :/ 

Thanks, I'm aware of that. I had to do some patching (not committed yet) [until it's resolved in ffigen](https://github.com/dart-lang/ffigen/issues/7#issuecomment-724627790)

vaind:
> It's +10,112 −1,483 LoC and I assume the generated code is most of that. Can you give a high level explanation why it's almost +9K net? Maybe some example?

Yes, most of it is the generated code, the (hopefully temporary) patch file & objectbox.h, 9777 lines added altogether.

image

The generated code is a lot larger then our manual code because it doesn't use templates (generics) which we used to reduce the amount of manual work (though it has made the code quite a bit less readable) and it is more type safe (all pointers now have types, like `Pointer<OBX_Store>` instead of the original `Pointer<Void>`. Also, it's actually complete (with the exception of `obx_cursor_*` which I've excluded for now (unused), as opposed to the original binding code where we've only defined what we needed. 

-----------------

> Could you point us to that patch you did?

https://github.com/objectbox/objectbox-dart/pull/146/files#diff-fdd27307bfa2cd39c097a0ca82b66dfce50d199482400073ea7b37399e74ff22

> Did you test this on a 32 bit system (in addition to 64 bit ofc); e.g. Android?

Yes, tested the example app on the 32-bit android. We can't run our complete test-suite on device [yet](https://github.com/objectbox/objectbox-dart/issues/81).

Issue #151, opened 2020-11-20
Stream of `query` notifies all boxes regardless of type


RTrackerDev:
Stream notifies all subscribers of changes no matter what box you listen.

I can offer a quick solution(PR: https://github.com/objectbox/objectbox-dart/pull/152):

```
void _setup() {
    if (!_Observable.isSubscribed(store)) {
      _Observable.subscribe(store);
    }
    final storeAddress = store.ptr.address;

    _Observable._any[storeAddress] ??= <int, Any>{};
    _Observable._any[storeAddress][entityId] ??= (u, pid, __) {
      // dummy value to trigger an event
      _Observable.controller.add(entityId);
    };
  }
```
I changed `_Observable.controller.add(pid)` -> `_Observable.controller.add(entityId)`

And:
```
Stream<List<T>> findStream({int offset = 0, int limit = 0}) {
    _setup();
    return _Observable.controller.stream.where((e)=>e == entityId)
        .map((_) => find(offset: offset, limit: limit));
  }

  /// Use this for Query Property
  Stream<Query<T>> get stream {
    _setup();
    return _Observable.controller.stream.where((e)=>e == entityId).map((_) => this);
  }
```
I added `.where((e)=>e == entityId)` to stream

PR #155, opened 2020-11-24
Fix Stream issue #151, with @RTrackerDev's suggestions

Buggaboo:
* add non-cross entity notification test
* made Query.store -> Query._store package private again
* one C observer callback per store, killed another bug
* add @RTrackerDev's fix


Buggaboo:
Crap. I didn't properly update my branch.

PR #162, opened 2020-12-31
ToOne Relations

### LINKED ISSUE #90: Relations
Am I missing something or does the dart lib not support relations yet?

vaind:
Part of #90 
* [x] adds `ToOne` property relation wrapper
* [x] ToOne query links

Side effects:
* expose `PutMode` enum, already a `box.put()` parameter but actually not available outside `box.dart`
* add `store.box<EntityT>()` factory

### DIFF
--- a/generator/lib/src/code_chunks.dart
+++ b/generator/lib/src/code_chunks.dart
@@ -126,6 +126,8 @@ class CodeChunks {
         case OBXPropertyType.Long:
         case OBXPropertyType.Date:
         case OBXPropertyType.DateNano:
+          fieldType = 'Integer';
+          break;
         case OBXPropertyType.Relation:
           fieldType = 'Relation';
           break;
--- a/generator/lib/src/code_chunks.dart
+++ b/generator/lib/src/code_chunks.dart
@@ -32,6 +32,7 @@ class CodeChunks {
     return """
       EntityDefinition<${name}>(
         model: model.getEntityByUid(${entity.id.uid}),
+        toOneRelations: ($name inst) => [${toOneRelationsList(entity).join(',')}],
         getId: ($name inst) => inst.${propertyFieldName(entity.idProperty)},
         setId: ($name inst, int id) {inst.${propertyFieldName(
         entity.idProperty)} = id;},
@@ -93,6 +94,12 @@ class CodeChunks {
       return "r.$field = members['${name}'];";
     }
   }
+
+  static List<String> toOneRelationsList(ModelEntity entity) =>
+      entity.properties
+          .where((ModelProperty prop) => prop.type == OBXPropertyType.Relation)
+          .map((ModelProperty prop) => "inst.${propertyFieldName(prop)}")
+          .toList(growable: false);
 
   static String _queryConditionBuilder(ModelEntity entity) {
     final ret = <String>[];
--- a/generator/lib/src/code_chunks.dart
+++ b/generator/lib/src/code_chunks.dart
@@ -18,8 +18,7 @@ class CodeChunks {
       final model = ModelInfo.fromMap(${JsonEncoder().convert(model.toMap())}, check: false);
       
       final bindings = <Type, EntityDefinition>{};
-      ${model.entities.map((entity) => "bindings[${entity
-      .name}] = ${entityBinding(entity)};").join("\n")} 
+      ${model.entities.map((entity) => "bindings[${entity.name}] = ${entityBinding(entity)};").join("\n")} 
       
       return ModelDefinition(model, bindings);
     }
@@ -34,8 +33,7 @@ class CodeChunks {
         model: model.getEntityByUid(${entity.id.uid}),
         toOneRelations: ($name inst) => [${toOneRelationsList(entity).join(',')}],
         getId: ($name inst) => inst.${propertyFieldName(entity.idProperty)},
-        setId: ($name inst, int id) {inst.${propertyFieldName(
-        entity.idProperty)} = id;},
+        setId: ($name inst, int id) {inst.${propertyFieldName(entity.idProperty)} = id;},
         reader: ($name inst) => {
           ${entity.properties.map(propertyReader).join(",\n")}
         },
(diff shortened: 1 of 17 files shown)


generator/lib/src/code_chunks.dart
```diff
@@ -58,22 +61,46 @@ class CodeChunks {
...
   static String propertyWriter(ModelProperty property) {
     final name = property.name;
+    final field = propertyFieldName(property);
     if (isTypedDataList(property)) {
-      return "r.${propertyFieldName(property)} = members['${name}'] == null ? null : ${property.dartFieldType}.fromList(members['${name}']);";
+      return "r.$field = members['${name}'] == null ? null : ${property
+          .dartFieldType}.fromList(members['${name}']);";
+    } else if (property.type == OBXPropertyType.Relation) {
+      return "r.$field.targetId = members['${name}'];" +
+          "\n r.$field.attach(store);";
```
greenrobot-team, line 92: If there are multiple relations, the store is also attached multiple times (e.g. see `RelatedEntityB`). Currently no side effects, so I guess this is fine?

lib/src/relations/to_one.dart
```diff
@@ -0,0 +1,126 @@
...
+
+  int get targetId => _value._id;
+
+  set targetId(int /*?*/ id) {
+    id ??= 0;
+    if (id == _value._id) return;
+    if (id == 0) {
+      _value = _ToOneValue<EntityT>.none();
+    } else if (_value._state == _ToOneState.unstored &&
+        id == _entity.getId(_value._object)) {
```
greenrobot-team, line 82: Can this case ever happen? Is _ToOneState.unstored not only set in the `target` setter on the pre-condition that the ID is 0?

lib/src/relations/to_one.dart
```diff
@@ -0,0 +1,126 @@
...
+      _value = _ToOneValue<EntityT>.stored(id, _value._object);
+    } else {
+      _value = _ToOneValue<EntityT>.lazy(id);
+    }
+  }
+
+  bool get hasValue => _value._state != _ToOneState.none;
+
+  void attach(Store store) {
```
greenrobot-team, line 91: Should there maybe be checks in places that use `_box` and `_entity` that throw a helpful exception, instructing to call `attach(store)`? Right now there will just be a null pointer exception if attach is not called, right (edit: yes)?

test/relations_test.dart
```diff
@@ -0,0 +1,146 @@
...
+
+    var read = env.box.get(1);
+    expect(read.tString, equals(src.tString));
+    expect(read.relA.hasValue, isTrue);
+    expect(read.relA.targetId, 1);
+    expect(read.relA.target.tInt, 42);
+    expect(read.relA.target.relB.target.tString, equals('B1'));
+
+    // attach an existing item
+    expect(read.relA.target.relB.target.relA.hasValue, isFalse);
```
greenrobot-team, line 48: Maybe refactor `read.relA.target.relB.target` into a variable? I have trouble reading this :)

lib/src/box.dart
```diff
@@ -9,8 +9,9 @@ import 'bindings/helpers.dart';
 import 'bindings/structs.dart';
 import 'modelinfo/index.dart';
 import 'query/query.dart';
+import 'relations/to_one.dart';
 
-enum _PutMode {
+enum PutMode {
```
greenrobot-team, line 14: So why is this now public again?

greenrobot-team (commented):
Review done, back to you @vaind!

generator/lib/src/code_chunks.dart
vaind: It's attached to each relation itself so I also don't see an issue.

lib/src/box.dart
vaind: hmm, I originally used it in to_one.dart (put was forwarded through there, instead of returning `internalTargetBox`). 
I could roll-back the change as it is unrelated to this PR, but I'd actually suggest keeping it anyway. We've already accepted mode param in `box.put()` but actually didn't allow users to pass any value :facepalm:

lib/src/relations/to_one.dart
vaind: happens here: https://github.com/objectbox/objectbox-dart/pull/162/files#diff-7f77c39ca497ed9b247c62f1f433e1d33f2ec7e03f0450b657b9d367584c0a0fR362 
```dart
void _putToOneRelFields(T object, PutMode mode) {
    _entity.toOneRelations(object).forEach((ToOne rel) {
      // put new objects
      if (rel.hasValue && rel.targetId == 0) {
        rel.targetId = rel.internalTargetBox._put(rel.target, mode, true);
      }
    });
  }
  ```
  
Or rather it will happen once put() actually calls `setId()` on new objects (that's already implemented in the optimizations PR #164 - https://github.com/objectbox/objectbox-dart/pull/164/files#diff-7f77c39ca497ed9b247c62f1f433e1d33f2ec7e03f0450b657b9d367584c0a0fR101)

File: lib/src/relations/to_one.dart (hunk 3 above)
Comment (vaind) [reply lib/src/relations/to_one.dart:91]: I was hoping to get this solved by null safety (late final) but yeah, let's not wait for that, I'll push shortly.

File: test/relations_test.dart (hunk 4 above)
Comment (vaind) [reply test/relations_test.dart:48]: OK

File: generator/lib/src/code_chunks.dart (hunk 1 above)
Comment (greenrobot-team) [reply generator/lib/src/code_chunks.dart:89]: 🤦  Uh, sorry... Not an issue.

File: lib/src/relations/to_one.dart (hunk 2 above)
Comment (greenrobot-team) [reply lib/src/relations/to_one.dart]: Ah, thanks, that makes sense. Maybe add a comment to refer to that put case (or put differently add the reason why this condition should exist)?

File: lib/src/relations/to_one.dart (hunk 3 above)
Comment (greenrobot-team) [reply lib/src/relations/to_one.dart:93]: LGTM!

File: test/relations_test.dart (hunk 4 above)
Comment (greenrobot-team) [reply test/relations_test.dart]: LGTM!

File: lib/src/box.dart (hunk 5 above)
Comment (greenrobot-team) [reply lib/src/box.dart:14]: Ah, ok. So users should be able to set this, then all good!

**Comment by greenrobot-team** [comment]
Except that suggestion to add a comment LGTM now. Back to @vaind.

File: lib/src/relations/to_one.dart (hunk 2 above)
Comment (vaind) [reply lib/src/relations/to_one.dart]: added a comment:
```
      // Optimization for targetId being set from box.put(sourceObject)
      // after entity.setId(object, newID) was already called on the new target.
```

PR #163, opened 2021-01-06
Performance benchmark

### LINKED ISSUE #144: Performance benchmark
We should have a set of benchmarks to check the performance of the current implementation and also to relate to when changes are made (for example when #68 is implemented). In short, the benchmark should address CRUD performance, bulk operations (`putMany()`) and queries.

You can take clue from other ObjectBox language bindings:
* [objectbox-go](https://github.com/objectbox/objectbox-go/tree/master/test/performance)
* [objectbox-java](https://github.com/objectbox/objectbox-performance)

Also, have a look at these blog posts:
* https://objectbox.io/how-to-benchmark-database-performance-and-objectbox/
* https://objectbox.io/go-1-0-release-and-performance-benchmarks/

vaind:
Part of #144


benchmark/bin/main.dart
```diff
@@ -0,0 +1,57 @@
+import 'dart:io';
```
greenrobot-team, line 1: As per https://dart.dev/tutorials/server/get-started#4-run-the-app I tried `dart run` in the `benchmark` folder. Fails with

```
C:\Users\uwe\greenrobot-git\objectbox-dart\benchmark>dart run
Could not find the implicit file to run: bin\benchmark.dart.
```

Should this file be `benchmark.dart` instead?

benchmark/lib/benchmark.dart
```diff
@@ -0,0 +1,84 @@
...
+      final fnTimes = times[fun];
+
+      final sum = fnTimes.map((d) => d.inMicroseconds).reduce((v, e) => v + e);
+      final avg = sum.toDouble() / fnTimes.length.toDouble() / 1000;
+      final timesCols = fnTimes.map((d) => d.inMicroseconds.toDouble() / 1000);
+      _print([fun, fnTimes.length, avg, ...timesCols]);
+    }
+  }
+
+  List<TestEntity> prepareData(int count) {
```
greenrobot-team, line 55: A little confused why `prepareData` and the operations (`putMany`, `updateAll`) are in this file and not e.g. `main.dart`. Should the measuring and the actual operations be separated, e.g. to allow easy re-use in the future?

greenrobot-team (commented):
Review done, not much to see here. Only had some organizational questions. Back to @vaind.

benchmark/bin/main.dart
vaind: OK, renamed

benchmark/lib/benchmark.dart
vaind: Actually, I wanted to have as much in the "lib" so that this can be also used to benchmark from a flutter application, without much code duplication.

vaind: I'll leave this discussion open and merge anyway, we can address directly on main later if needed.

greenrobot-team: It's fine, just wanted to point it out. Can refactor when the time comes.

Issue #68, opened 2019-10-31
Flatbuffers marshalling performance


vaind:
The way we currently marshal and unmarshal flatbuffers into objects isn't optimal. We should get rid of the intermediary `propVals map<string, dynamic>`. Ideally creating objects (or flatbuffers in the other direction) directly in the generated code.

greenrobot:
Yes, this should use code generation for both ways.

Once this is done and reviewed, please give me the final review.

Buggaboo:
Can we borrow anything from [google's flatc](https://github.com/google/flatbuffers/blob/master/src/idl_gen_dart.cpp)?

greenrobot:
We should have a minimal CRUD perf app before doing that?

Buggaboo:
What is the bottleneck here precisely? Runtime property selection using the Map? Can we use the offsets to get/put the data from a stream of bytes?

greenrobot:
Look at how Go does it, e.g. the generated code here:

- [Flatten()](https://github.com/objectbox/objectbox-go/blob/5bc8560255c2501c47feaeef8c1e4d61aada2fc1/test/generator/testdata/task/byvalue.obx.go.expected#L91):  object (struct) to FlatBuffers
- [Load()](https://github.com/objectbox/objectbox-go/blob/5bc8560255c2501c47feaeef8c1e4d61aada2fc1/test/generator/testdata/task/byvalue.obx.go.expected#L102): FlatBuffers to object

Go does the FlatBuffers handling directly, while Java FlatBuffers handling is to slow and defers that to JNI. I hope compiled Dart is closer to Go performance, so it's at least worth a try.

Buggaboo:
It could be even faster with caching probably e.g. maintain _last_edited_ and _last_read_ timestamps in atomic transactions.

greenrobot:
ObjectBox already benefits from OS cached memory mapping (see [FAQ](https://objectbox.io/faq/) for some details). We did not bother about object-level caches so far, as ObjectBox gets like 1 million objects per second from the DB on regular notebooks.

Buggaboo:
I'm looking into [benchmark harness(https://pub.dev/packages/benchmark_harness).

How does one stress test the library? _N_ number of writes to a random (sub)set of fields?

vaind:
[...]

PR #164, opened 2021-01-06
Optimizations

### LINKED ISSUE #62: Box insert - assign new IDs on passed objects
The passed object should be updated to also contain ID after a successful insert. This applies to `box.put()` and `box.putMany()`

Example for `put()`:
```dart
final object = TestEntity.initText("Hello");
int putId = box.put(object);
expect(putId, 1);
expect(object.id, 1);
```

vaind:
Tries to improve performance as evaluated by #144. Based on PR #163 - should be merged first. Also closes #62 

* [x] using [custom flatbuffers builder](https://github.com/objectbox/objectbox-dart/pull/164/commits/cc876475a195451cc9b5d44f604d01280c4653fe) with c-backed storage: putMany 110 ms -> 50 ms
* [x] [reusing flatbuffers builder & switching to cursor](https://github.com/objectbox/objectbox-dart/pull/164/commits/156457944f7789837b8c17324ee7e3b9139066f8) further improved putMany 50 ms -> 45 ms
* [x] [use FFI `memset()` to clear flatbuffers buffer on init/reset](https://github.com/objectbox/objectbox-dart/pull/164/commits/a96f6732b94ebfd91146a04a0dc03e91495209a2) instead of dart-only alternatives - putMany 45ms -> 40 ms
* [x] [group together value collection and flatbuffers marshalling](https://github.com/objectbox/objectbox-dart/pull/164/commits/534523cf7e1da333ff54526088df45a502093807) - putMany 40 ms -> 35 ms
* [x] use data visitors & generated unmarshalling code - readAll() 50 ms -> 10 ms


lib/flatbuffers/flat_buffers.dart
```diff
@@ -126,17 +130,23 @@ class Builder {
 
   void _bufClear([int length]) {
     length ??= _buf.lengthInBytes;
-    // TODO these are slow, can we use memset() from DynamicLibrary.process()?
+    // Note: memset is faster than any of the following alternatives
+    _memset ??= DynamicLibrary.process().lookupFunction<
```
greenrobot, line 134: What's the clearing for? E.g. you could zero out the first 4 bytes, where the pointer to FB content resides.

vaind: 1. makes sure to produce the same output as the upstream flatbuffers (see flatbuffers_test.dart which otherwise fails). Of course I can remove that test if you're sure data can be uninitialized... 
2. apparently it has some other impacts. Without the clear, of with just clearing the first 4 bytes as you suggested, some tests fail suddenly (queries on strings). Which made me have another look and I believe there's another bug in dart flatbuffers implementation, let's see... https://github.com/google/flatbuffers/pull/6390
3. Valgrind complains: `Conditional jump or move depends on uninitialised value(s)`... don't know where, though...
```
==355627== Conditional jump or move depends on uninitialised value(s)
==355627==    at 0x504A4E8: ??? (in /tmp/t)
==355627==    by 0x4FD245D: ??? (in /tmp/t)
==355627==    by 0x4FE19DF: ??? (in /tmp/t)
==355627==    by 0x4FE186B: ??? (in /tmp/t)
==355627==    by 0x5049DE4: ??? (in /tmp/t)
==355627==    by 0x506A6B5: ??? (in /tmp/t)
==355627==    by 0x506C4DC: ??? (in /tmp/t)
==355627==    by 0x5064464: ??? (in /tmp/t)
==355627==    by 0x5053268: ??? (in /tmp/t)
==355627==    by 0x5037B89: ??? (in /tmp/t)
==355627==    by 0x505339D: ??? (in /tmp/t)
==355627==    by 0x500C816: ??? (in /tmp/t)
```

greenrobot: What's the clearing for, in a sense of what is it needed for? (the purpose on it's own, regardless of different things)

vaind: I don't get the question. I've answered to the best of my knowledge. I have no idea if/why flatbuffers internally need the buffer cleared. It may be the case they're relying on the values being zero (as with the linked PR), it may be not necessary at all :man_shrugging:

greenrobot: Do you want to do the same as C++`s FlatBufferBuilder.Clear()? But you do not have the FlatBufferBuilder? Can discuss tomorrow in vcall if that is getting too involved.

vaind: this code **is** inside FlatBuffer Builder. C++ FBB's `clear()` is called `reset()` in dart's FBB. 

Yep, let's address in a call, seems like we aren't on the same page.

greenrobot: Lolz, indeed I did not notice that we are discussing (official?) FlatBuffers code here. C++ zeroes out a few bytes, right?

vaind: conversation moved

lib/flatbuffers/flat_buffers.dart
```diff
@@ -0,0 +1,1317 @@
+// Copyright (c) 2016, the Dart project authors.  Please see the AUTHORS file
```
greenrobot-team, line 1: As said, it would be nice to see the actual changes to the original flat_buffers.dart (also possibly makes rebasing these changes onto newer upstream flatbuffers versions easier).

greenrobot-team: Flatbuffers PR for Allocator interface: https://github.com/google/flatbuffers/pull/6397

lib/src/box.dart
```diff
@@ -163,75 +121,73 @@ class Box<T> {
...
+      } finally {
+        free(dataPtrPtr);
+        free(sizePtr);
+        checkObx(bindings.obx_cursor_close(cCursor));
+      }
+    });
   }
 
   /// Returns all stored objects in this Box.
   List<T> getAll() {
```
greenrobot-team, line 164: Maybe document that `getAll()` and maybe `getMany()` now return fixed-size lists (as the default is growable, no?)?

E.g. imagine a use case like: get and then modify some objects and add some new objects, then calling putMany. That wouldn't work now.

greenrobot-team:
Side note: Was on a goose chase because in the generated `objectToFB` code the field ID for `tString` looked fishy (tests are all good though!). Turns out that in `test/objectbox-model.json` there are no field/property IDs 2 to 5. 🤦

generator/lib/src/code_chunks.dart
```diff
@@ -33,45 +36,109 @@ class CodeChunks {
...
     return property.name;
   }
 
-  static String propertyReader(ModelProperty property) {
-    return "'${property.name}': inst.${propertyFieldName(property)}";
+  static int propertyFlatBuffersSlot(ModelProperty property) =>
+      property.id.id - 1;
+
+  static int propertyFlatBuffersvTableOffset(ModelProperty property) =>
+      4 + 2 * propertyFlatBuffersSlot(property);
```
greenrobot-team, line 53: Where does this calculation come from?

greenrobot-team (commented):
Review done. Tried to make sense of this as good as possible (e.g. can't verify if C lib cursor API used correctly as I never used it). Back to @vaind with a suggestion and questions.

lib/flatbuffers/flat_buffers.dart
vaind: yep, that's exactly what's changed. The final version of the code here is an exact copy of the code in the PR

generator/lib/src/code_chunks.dart
vaind: ehh, I don't have any docs reference. It's just something I've found out from the flatc generated code and use in all the language bindings. e.g. [c++](https://github.com/objectbox/objectbox-generator/blob/main/test/comparison/testdata/fbs/typeful/cpp/schema.obx.cpp.expected#L36)

lib/src/box.dart
vaind: getAll actually returns a growable list (we don't know the size beforehand). 

However, you've inspired me to change getMany(), adding a `growableResult=false` named optional param, allowing user to pick whether they need the (slower) growable list

vaind:
@greenrobot [back in 2019 you wanted to have a look before a merge](https://github.com/objectbox/objectbox-dart/issues/68#issuecomment-548369843)

generator/lib/src/code_chunks.dart
greenrobot-team: K, so I'll just believe this works then. 😁

lib/src/box.dart
greenrobot-team: `getAll()` is fine, not sure what I was seeing.

`getMany()` growable change: https://github.com/objectbox/objectbox-dart/commit/8a551e49c44e0192ca59abd39d148bb8274a1d80 LGTM!

PR #170, opened 2021-01-15
Rework readme and example/readme

vaind:
Closes #135 

I've basically rewritten both README.md files so I suggest not looking at the changes but rather at final files themselves.

PR #171, opened 2021-01-16
make @Id annotation optional - recognize automatically

### LINKED ISSUE #167: Recognize id field without an annotation
It should be very simple to support `id` fields without annotation, see https://github.com/objectbox/objectbox-dart/blob/main/generator/integration-tests/basics/lib/lib.dart#L18

If there is no field annotated `Id()`, try to find one:
* there must be a single field called "ID" (case insensitive), and its type must be `int`

vaind:
closes #167

### DIFF
--- a/benchmark/lib/model.dart
+++ b/benchmark/lib/model.dart
@@ -2,7 +2,6 @@ import 'package:objectbox/objectbox.dart';
 
 @Entity()
 class TestEntity {
-  @Id()
   int /*?*/ id;
 
   String /*?*/ tString;
--- a/example/flutter/objectbox_demo/lib/main.dart
+++ b/example/flutter/objectbox_demo/lib/main.dart
@@ -11,7 +11,6 @@ import 'objectbox.g.dart';
 
 @Entity()
 class Note {
-  @Id()
   int id;
 
   String text;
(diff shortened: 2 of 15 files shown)


example/README.md
```diff
@@ -45,7 +45,6 @@ import "package:objectbox/objectbox.dart";
 
 @Entity()
 class Note {
-    @Id()       // required; stored as a 64-bit unsigned integer in ObjectBox
```
greenrobot-team, line 48: Should the README point out that using `@Id` is possible (e.g. doesn't this give the impression field must be named `id`)? Also that there is a naming convention if not using it?

greenrobot-team (changes requested):
Review done.

Should there at least be one usage/test remaining to ensure `@Id` works?

example/README.md
greenrobot-team: Never mind, #170 fixes this.

greenrobot-team:
> You mean like this? https://github.com/objectbox/objectbox-dart/pull/171/files#diff-c4a35df46b0e307508b4ef78832a01ca3d549cb0139ed2fb275300ff5fc1fe21R17

Find usages in IDEA did totally not find that, sorry. All good!

PR #175, opened 2021-01-21
Box - change constructor to factory, reuse a single FlatBuffers builder instance

vaind:


greenrobot-team (commented):
LGTM. But isn't (some) entity data now kept around in memory for each Box until a Store is closed? E.g. memory usage increases with each new (used) Box instance, no?

vaind:
> memory usage increases with each new (used) Box instance

Correct, but there's only a single instance per entity, so I figured that should be fine?

greenrobot-team:
> > memory usage increases with each new (used) Box instance
> 
> Correct, but there's only a single instance per entity (type), so I figured that should be fine?

Just wanted to mention it figuring esp. on mobile memory usage is a concern.

vaind:
> keep the buffer if below 64 KB 

done and merged

PR #176, opened 2021-01-25
Multi-platform & multi-version CI

### LINKED ISSUE #80: CI on macOS and Windows
Extend current CI (GH Actions) to also run on macOS and Windows.

### LINKED ISSUE #82: CI - test all supported dart/flutter versions
Currently, we use `image:  google/dart:latest` in our GitHub. Maybe we should instead explicitly test all versions we currently support (e.g. 2.6 and 2.7).

vaind:
* closes #80 
* closes #82

### DIFF
--- a/lib/src/bindings/flatbuffers.dart
+++ b/lib/src/bindings/flatbuffers.dart
@@ -82,15 +82,28 @@ class _Allocator extends fb.Allocator {
   @override
   void clear(ByteData data, bool _) {
     if (_memset == null) {
-      DynamicLibrary lib;
-      if (Platform.isWindows) {
-        // DynamicLibrary.process() is not available on Windows
-        lib = DynamicLibrary.open('msvcr100.dll');
-      } else {
-        lib = DynamicLibrary.process();
+      try {
+        DynamicLibrary lib;
+        if (Platform.isWindows) {
+          // DynamicLibrary.process() is not available on Windows, let's load a
+          // lib that defines 'memset()' it - should be mscvr100 or mscvrt DLL.
+          // mscvr100.dll is in the frequently installed MSVC Redistributable.
+          lib = DynamicLibrary.open('msvcr100.dll');
+        } else {
+          lib = DynamicLibrary.process();
+        }
+        _memset = lib.lookupFunction<
+            Void Function(Pointer<Void>, Int32, IntPtr),
+            _dart_memset>('memset');
+      } catch (_) {
+        // fall back if we can't load a native memset()
+        _memset = (Pointer<Void> ptr, int byte, int size) {
+          final bytes = ptr.cast<Uint8>();
+          for (var i = 0; i < size; i++) {
+            bytes[i] = byte;
+          }
+        };
       }
-      _memset = lib.lookupFunction<Void Function(Pointer<Void>, Int32, IntPtr),
-          _dart_memset>('memset');
     }
     _memset(_allocs[data].cast<Void>(), 0, data.lengthInBytes);
   }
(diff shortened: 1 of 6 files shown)


lib/src/bindings/flatbuffers.dart
```diff
@@ -80,8 +81,17 @@ class _Allocator extends fb.Allocator {
 
   @override
   void clear(ByteData data, bool _) {
-    _memset ??= DynamicLibrary.process().lookupFunction<
-        Void Function(Pointer<Void>, Int32, IntPtr), _dart_memset>('memset');
+    if (_memset == null) {
+      DynamicLibrary lib;
+      if (Platform.isWindows) {
+        // DynamicLibrary.process() is not available on Windows
+        lib = DynamicLibrary.open('msvcr100.dll');
```
greenrobot-team, line 88: Where is this workaround coming from? Appears to be a C library. Does this require installation of Microsoft Visual C++ 2015 Redistributable (x64) like mentioned at https://docs.objectbox.io/java-desktop-apps#native-libraries? Should this be noted in a README?

The GitHub Windows instance has a bunch of versions installed, so I guess most are fine. https://github.com/actions/virtual-environments/blob/win19/20201210.0/images/win/Windows2019-Readme.md#microsoft-visual-c

lib/src/query/query.dart
```diff
@@ -754,7 +754,7 @@ class Query<T> {
     try {
       final idArray = idArrayPtr.ref;
       return idArray.count == 0
-          ? List<int>.empty()
+          ? List<int>.filled(0, 0)
```
greenrobot-team, line 757: Did this sneak in while trying to fix https://github.com/objectbox/objectbox-dart/issues/177?

pubspec.yaml
```diff
@@ -20,8 +21,8 @@ dev_dependencies:
   build_runner: ^1.0.0
   objectbox_generator:
     path: generator
-  pedantic: ^1.10.0-nullsafety.0
-  test: ^1.16.0-nullsafety.0
+  pedantic: ^1.9.0
+  test: ^1.0.0
```
greenrobot-team, line 25: I guess this is unrelated? But fine.

tool/pub.sh
```diff
@@ -0,0 +1,12 @@
...
+set -euo pipefail
+
+# forward all arguments to an available version of `pub`
+
+if [[ `command -v pub` ]]; then
+  pub "$@"
+elif [[ `command -v pub.bat` ]]; then
+  pub.bat "$@"
+else
+  dart pub "$@"
```
greenrobot-team, line 11: Is the `dart` command not supposed to replace the above in the future? E.g. should this be checked first? But I guess fine (dart docs do not seem to indicate any difference https://dart.dev/tools/dart-tool).

greenrobot-team (commented):
LGTM, just some questions.

lib/src/query/query.dart
vaind: Nope, that's to restore compatibility with Dart 2.7

pubspec.yaml
vaind: nope, that's to restore compatibility with anything prior to 2.10

tool/pub.sh
vaind: It has replaced `pub`, but this check is simpler than trying to see what happens if you call `dart pub` on an old SDK (it fails)

lib/src/bindings/flatbuffers.dart
vaind: > Where is this workaround coming from?

Nowhere in particular, I've just searched what libary defines `memset` on windows. Turns out it's `mscvrt.dll` or `mscvr100.dll`. Yes it may be fragile if users don't have the lib... I'll add another fallback layer

greenrobot-team: Maybe add this in a comment then?

vaind: OK, also added a comment

PR #178, opened 2021-01-25
Flutter Desktop demo

vaind:
Updates the existing demo to work on Flutter desktop

In short, I've: 
*  run `flutter create .` in the demo folder
* removed `example/flutter/objectbox_demo_desktop`

PR #179, opened 2021-01-25
Move objectbox package sources to a separate directory

### LINKED ISSUE #165: Move objectbox package to its own folder
We currently have [multiple packages published pub.dev](https://pub.dev/publishers/objectbox.io/packages), all of them coming from this repository. However, with objectbox being published directly from the root folder, the published package contains everything in the repository, [including all the other packages](https://storage.googleapis.com/pub-packages/packages/objectbox-0.10.0.tar.gz). While this doesn't influence the compiled app size, it would be cleaner if we moved objectbox into its own folder as well, keeping the root of the repo only for scripts, tools, etc.

Additionally, it would allow us to have different README.md for the pub.dev front-page and for the github repo (without having to use scripting when publishing.

vaind:
Closes #165 

Pretty straightforward move with a couple of path changes (mainly CI). I've also tried updating publishing scripts but didn't run them so there may be some more changes necessary during the next release.

Wasn't sure what the root README.md should look like, so didn't create one yet.


greenrobot-team:
The root README.md kind of should be the same, no? Maybe move it back and copy it to the new subfolder before publishing?

.github/workflows/dart.yml
```diff
@@ -44,11 +44,15 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install ObjectBox C-API
         run: ./install.sh
-      - run: ./tool/pub.sh get
-      - name: Generate ObjectBox models
-        run: ./tool/pub.sh run build_runner build
-      - name: Run tests
-        run: ./tool/pub.sh run test
+      - if: startsWith(matrix.os, 'windows')
+        run: mv lib/objectbox.dll objectbox/lib/
```
greenrobot-team, line 48: Maybe run `install.sh` from the subfolder? Or maybe move it there or to the `tool` folder?

greenrobot-team (commented):
Script changes and `pub publish --dry-run` LGTM. I'd say a GitHub main README should be added before merging, it's the first thing existing users see when coming here.

Note: this conflicts with #178

.github/workflows/dart.yml
vaind: yeah, I can run it from the subfolder... Won't move because we link from it in the installation instructions:

bash <(curl -s https://raw.githubusercontent.com/objectbox/objectbox-dart/main/install.sh)

greenrobot-team: Oops, didn't look closely, thought it was the objectbox-c install script...

vaind:
> README

I've added a symbolic link for now, if we decide to make some changes, we can add a regular file later

PR #180, opened 2021-01-26
Avoid recursive runInTransaction calls with relations

vaind:


objectbox/lib/src/transaction.dart
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  /// Executes a given function inside a transaction.
+  ///
+  /// Returns type of [fn] if [return] is called in [fn].
+  static R execute<R>(Store store, TxMode mode, R Function() fn) {
+    final tx = Transaction(store, mode);
+    try {
+      if (tx._isWrite) tx.markSuccessful(true);
+      return fn();
```
greenrobot-team, line 76: I know this is the same behavior as before, but shouldn't the transaction be marked successful *after* `fn` executed? E.g. similar to `put()`?

objectbox/lib/src/relations/to_many.dart
```diff
@@ -201,7 +204,10 @@ class ToMany<EntityT> extends Object with ListMixin<EntityT> {
             throw UnimplementedError();
         }
       });
-    });
+      tx.markSuccessful(true);
```
greenrobot-team, line 207: Also `if (ownedTx)`? Again, I'm not sure how markSuccessful works with ObjectBox.

greenrobot-team (commented):
Some questions, otherwise LGTM.

objectbox/lib/src/transaction.dart
vaind: Yes, in the perfect world where we wouldn't have to define a variable storing the result of `fn()` before returning it after a `markSuccessful(true)` call. This is more efficient, assuming transactions "usually" don't fail

objectbox/lib/src/relations/to_many.dart
vaind: you can call it multiple times, but yes, no need to do it here if not owned.

objectbox/lib/src/transaction.dart
greenrobot-team: So this is intended, maybe add a note?

vaind: OK

PR #181, opened 2021-01-28
Run tests as a standalone flutter app

vaind:
Currently, all supported functionality is tested in dart, with tests in `objectbox/test/`.
While this should work fine and is great for development, it would be nice if we could run the same test suite (or at 
least most of it), in a Flutter application (test), making sure everything works on all platforms. 
This is especially important given the C-API version provided for each platform may differ slightly (released at different times).

My idea is to have a look if I can load those tests somehow from a flutter app/flutter test.

Note: the first commit can be ignored for review. It's a result of `flutter create integration_test`


vaind:
This is too outdated to continue working on, closing.

PR #183, opened 2021-02-01
API & docs changes (post merge review)

vaind:


generator/lib/src/entity_resolver.dart
```diff
@@ -153,7 +153,9 @@ class EntityResolver extends Builder {
               "  invalid relation property '${f.name}' in entity '${element.name}' - must use ToOne/ToMany<TargetEntity>");
           continue;
         }
-        relTargetName = (f.type as ParameterizedType).typeArguments[0].name;
+        relTargetName = (f.type as ParameterizedType)
+            .typeArguments[0]
+            .getDisplayString(withNullability: false);
```
greenrobot-team, line 158: The docs say "Clients should not depend on the content of the returned value as it will be changed if doing so would improve the UX.". But this string is used in e.g. the code generator, no? Is this a good idea then?

objectbox/lib/src/store.dart
```diff
@@ -93,35 +101,66 @@ class Store {
...
   SyncClient /*?*/ syncClient() => syncClientsStorage[this];
 
   /// The low-level pointer to this store.
   Pointer<OBX_store> get ptr => _cStore;
 }
+
+/// Internal only.
+// TODO enable annotation once meta:1.3.0 is out
+// @internal
+class InternalStoreAccess {
```
greenrobot-team, line 154: These internal access classes still make me cringe. Shouldn't public interfaces and internal implementations hide API from users?

objectbox/lib/src/annotations.dart
```diff
@@ -1,27 +1,57 @@
+/// Entity annotation is used to on a class to let ObjectBox know it should
```
greenrobot-team, line 1:
```suggestion
/// Entity annotation is used on a class to let ObjectBox know it should
```

objectbox/lib/src/annotations.dart
```diff
@@ -1,27 +1,57 @@
...
   const Entity({this.uid});
 }
 
-/// A dart int value can map to different OBXPropertyTypes,
-/// e.g. Short (Int16), Int (Int32), Long (Int64), all signed values.
-/// Also a dart double can also map to e.g. Float and Double
-///
-/// Property allows the mapping to be specific. The defaults are
-/// e.g. Int -> Int64, double -> Float64, bool -> Bool.
+/// Property annotation enables you to explicitly specify configure some details
```
greenrobot-team, line 24:
```suggestion
/// Property annotation enables you to explicitly configure some details
```

objectbox/lib/src/annotations.dart
```diff
@@ -1,27 +1,57 @@
...
+/// about how a field is stored in the database.
 ///
-/// Use OBXPropertyType and OBXPropertyFlag values, resp. for type and flag.
+/// Use PropertyType and OBXPropertyFlag values, resp. for type and flag.
```
greenrobot-team, line 27:
```suggestion
/// Use PropertyType values for type.
```
Or remove altogether?

objectbox/lib/src/annotations.dart
```diff
@@ -1,27 +1,57 @@
...
+  /// database accordingly on the next application launch - renaming the stored
+  /// Property instead of creating a new one.
+  final int /*?*/ uid;
 
-  /// Override dart type with an alternative ObjectBox property type
+  /// Override dart type with an alternative ObjectBox property type.
+  ///
+  /// A dart int value can map to different [PropertyType]s,
+  /// e.g. Short (Int16), Int (Int32), Long (Int64), all signed values.
+  /// Also a dart double can also map to e.g. Float and Double
```
greenrobot-team, line 45:
```suggestion
  /// Also a dart double can map to e.g. Float and Double
```

objectbox/lib/src/annotations.dart
```diff
@@ -74,11 +104,16 @@ enum PropertyType {
   // stringVector
 }
 
+/// Annotation Id can be used to specify an entity ID property if it's named
+/// anything else then "id" (case insensitive).
```
greenrobot-team, line 108:
```suggestion
/// anything else than "id" (case insensitive).
```
But "named anything else than" sounds wrong to me anyhow. Not sure how to say this properly though.

objectbox/lib/src/annotations.dart
```diff
@@ -74,11 +104,16 @@ enum PropertyType {
...
 class Id {
+  /// Create an Id annotation.
   const Id();
 }
 
+/// Transient annotation marks properties that should not be stored in the DB.
```
greenrobot-team, line 114:
```suggestion
/// Transient annotation marks properties that should not be stored in the database.
```
Maybe?

objectbox/lib/src/annotations.dart
```diff
@@ -97,30 +132,56 @@ class Transient {
 /// Note: indexes are currently not supported for ByteVector, Float or Double
 /// properties.
 class Index {
+  /// Index type.
   final IndexType /*?*/ type;
 
+  /// Create an Index annotaion.
```
greenrobot-team, line 138:
```suggestion
  /// Create an Index annotation.
```

objectbox/lib/src/annotations.dart
```diff
@@ -97,30 +132,56 @@ class Transient {
...
+/// [value] indexes instead for [String] properties.
 enum IndexType {
+  /// Uses property values to build the index. This is a default for scalar
+  /// properties.
+  ///
+  /// If used for a [String] property, this may require more storage space than
+  /// it's default, [hash].
   value,
+
+  /// Uses a 32-bit hash of the field value to build the index. This is a
```
greenrobot-team, line 156:
```suggestion
  /// Uses a 32-bit hash of the field value to build the index. This is the
```
?

objectbox/lib/src/observable.dart
```diff
@@ -61,20 +62,25 @@ class _Observable {
       _Observable._anyObserver.containsKey(store.ptr.address);
 }
 
+/// Streamable adds stream support to queries. The stream reruns the query
+/// whenever there's a change in any of the object in the queried Entity
```
greenrobot-team, line 66:
```suggestion
/// whenever there's a change in any of the objects in the queried Entity
```
Also maybe `Box` instead of `Entity`? Then also below for `findStream()`.

objectbox/lib/src/relations/to_one.dart
```diff
@@ -109,13 +115,20 @@ class ToOne<EntityT> {
     }
   }
 
+  /// Whether the relatin field has a value stored. Otherwise it's null.
```
greenrobot-team, line 118:
```suggestion
  /// Whether the relation field has a value stored. Otherwise it's null.
```

greenrobot-team (commented):
Reviewed as good as possible (too many changes). Mostly found typos. But the display string change looks fishy.

objectbox/lib/src/store.dart
vaind: Yeah, dislike it too, but couldn't think of a better way to have some "package-protected" access among `lib/src/*.dart` files. The problem is, once some identifier starts with an underscore, you can only access it from the same file, because in Dart, file is a library. Of course you can also use `part` instead of import, basically doing the same thing - turning everything into a single library...

objectbox/lib/src/annotations.dart
vaind: yes, removed

generator/lib/src/entity_resolver.dart
vaind: `.name` got deprecated... I misread the suggestion, I was supposed to use *either* `element` or `getDisplayString()`... `element` is the way to go so changed it.

vaind:
Thanks for the review, good eye for those typos.

objectbox/lib/src/store.dart
greenrobot-team: As I said, isn't that what interfaces are for? E.g. make `Box` an abstract class (as Dart apparently does not have explicit interfaces) and create `BoxImpl` as the actual implementation. The implementation can have any number of public methods that users (which work with the interface) do not see (unless they use the implementation class directly, which docs should discourage).

Issue #185, opened 2021-02-01
Web-capable Box API; based on IndexedDB


greenrobot:
To support the Flutter Web platform, we need an alternative to native libraries, as wasm does not seem to cover everything the native libs do.

To provide a minimal valuable implementation, implement the Box API using IndexedDB (and FlatBuffers similar to how we use it in the native side). No relations, indexes etc.

Once that works, open a ticket to do simple queries next.

vaind:
Yes, that shouldn't be too hard to do. I've already thought about this a little a while back and how I've seen this done in Dart is:
* extract interfaces of the public APIs (Store, Box for now) instead of concrete classes
* have an implementation of those interfaces depending on the platform using [conditional imports](https://dart.dev/guides/libraries/create-library-packages#conditionally-importing-and-exporting-library-files)

vaind:
Also a thing to consider: not all browsers [support IndexedDB](https://api.dart.dev/stable/2.10.5/dart-indexed_db/dart-indexed_db-library.html), so if we wanted to have a complete solution, we'll need (to do) something like https://pub.dev/packages/idb_shim

greenrobot:
With like 97% of browsers supporting it we might be fine.

Another thing to consider: if we'd have the lower-level implemented in Type-/JavaScript, we could use that for plain web apps too. That probably makes sense when we start adding more features like queries.

vaind:
Since this would require API changes (switch to interfaces) we should try to prepare those prior to 1.0, with the actual implementation coming later as a minor release.

vaind:
> if we'd have the lower-level implemented in Type-/JavaScript, we could use that for plain web apps too.  

notes: 
* to implement interaction with JavaScript (i.e. call a objectbox-js impelmentation from Dart) there's `dart:js` and `package:js`, looks like the latter supersedes the former. Regarding the performance of such an interop - couldn't find anything useful yet. From my understanding, there shouldn't be much cost to calling javascript, considering the rest of the dart code is compiled to javascript as well, maybe just some checks/data validations before crossing the boundary to the code generated by dart2js.
* there's no support to integrate dart with TypeScript (maybe except if you compiled your TS to JS but I don't know how that works), see this ancient [Dart SDK issue](https://github.com/dart-lang/sdk/issues/20215).
[...]

PR #189, opened 2021-02-05
Web support vol.1

vaind:
Part of #185

* Shuffles stuff around and uses conditional exports to add "support" JS/web. 
* Actual support not implemented yet, just won't import anything not available on JS/web platforms, thus the analyzer thinks package supports it.
* **I've done my best to make this reviewable commit by commit. Ignore "Files changed" tab in the PR, it doesn't show up the moved files properly.**

### TODO:
* [ ] Don't merge as is, we should actually add some import that breaks web support before releasing, to avoid user confusion. E.g. add the following code to `lib/objectbox.dart`: `import 'dart:ffi';`
* [ ] ALso for the same reason don't merge the CI commit, it's here just so that we don't break the web support unintentionally when changes are made before merging


greenrobot-team:
Note: Dart docs about conditional imports https://dart.dev/guides/libraries/create-library-packages#conditionally-importing-and-exporting-library-files

vaind:
> Note: Dart docs about conditional imports https://dart.dev/guides/libraries/create-library-packages#conditionally-importing-and-exporting-library-files

Make sure to read the linked issue in case you haven't (the link's there)

greenrobot-team:
> Make sure to read the linked issue in case you haven't (the link's there)

Thanks! Only read replies today, so missed it...

greenrobot-team (commented):
LGTM, no obvious issues, still builds and tests run on my Windows setup.

I'd imagine not having an "interface" or stub implementation could lead to deviations in the two APIs. Does the compiler somehow check the APIs match between implementations? But I guess tests would uncover this quickly.

Also having to copy some of the generated constants is a step back, but I don't see another solution as well.

PR #223, opened 2021-04-22
Exceptions

### LINKED ISSUE #182: Request for comments: Unify exception throwing
We're currently throwing `ObjectBoxException` as well as `Exception`. 

* Either get rid of the former or use it in all places? 
* Or actually use more specific classes, depending on what kind of exception it is? Like `ArgumentError`, `StateError` and reserve the generic `Exception` for cases when we can't do any better?
* What's the best practice?

vaind:
closes #182

### DIFF
--- a/objectbox/lib/src/modelinfo/modelproperty.dart
+++ b/objectbox/lib/src/modelinfo/modelproperty.dart
@@ -23,7 +23,7 @@ class ModelProperty {
 
   set name(String? value) {
     if (value == null || value.isEmpty) {
-      throw Exception('name must not be null or an empty string');
+      throw ArgumentError('name must not be null or an empty string');
     }
     _name = value;
   }
@@ -32,7 +32,7 @@ class ModelProperty {
 
   set type(int? value) {
     if (value == null || value < 0) {
-      throw Exception('type must be defined and may not be < 0');
+      throw ArgumentError('type must be defined and may not be < 0');
     }
     _type = value;
   }
@@ -41,7 +41,7 @@ class ModelProperty {
 
   set flags(int? value) {
     if (value == null || value < 0) {
-      throw Exception('flags must be defined and may not be < 0');
+      throw ArgumentError('flags must be defined and may not be < 0');
     }
     _flags = value;
   }
@@ -61,7 +61,7 @@ class ModelProperty {
   set indexId(IdUid? value) {
     if (value != null) {
       if (value.id == 0 || value.uid == 0) {
-        throw Exception('indexId must contain valid ID & UID');
+        throw ArgumentError('indexId must contain valid ID & UID');
       }
     }
     _indexId = value;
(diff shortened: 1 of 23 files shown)


objectbox/lib/src/modelinfo/modelproperty.dart
```diff
@@ -23,7 +23,7 @@ class ModelProperty {
 
   set name(String? value) {
     if (value == null || value.isEmpty) {
-      throw Exception('name must not be null or an empty string');
+      throw ArgumentError('name must not be null or an empty string');
```
greenrobot-team, line 26: Maybe use `ArgumentError.value(...)` in this file, too? But I guess it just affects formatting of the output and does not provide source code linking or similar.

objectbox/lib/src/native/sync.dart
```diff
@@ -554,7 +554,7 @@ class Sync {
   /// Create a Sync annotation, enabling synchronization for an entity.
   const Sync();
 
-  static late final bool _syncAvailable = C.sync_available();
+  static late final bool _syncAvailable = C.has_feature(OBXFeature.Sync);
```
greenrobot-team, line 557: I guess this just snuck in here? Fine to keep, just double checking if not by accident.

objectbox/lib/src/relations/to_many.dart
```diff
@@ -243,7 +244,7 @@ class ToMany<EntityT> extends Object with ListMixin<EntityT> {
 
   void _verifyAttached() {
     if (!_attached) {
-      throw Exception('ToMany relation field not initialized. '
+      throw ObjectBoxException('ToMany relation field not initialized. '
```
greenrobot-team, line 247: Why not `StateError`?

objectbox/lib/src/relations/to_one.dart
```diff
@@ -142,7 +143,7 @@ class ToOne<EntityT> {
 
   void _verifyAttached() {
     if (!_attached) {
-      throw Exception('ToOne relation field not initialized. '
+      throw ObjectBoxException('ToOne relation field not initialized. '
```
greenrobot-team, line 146: Same as to-many: why not `StateError`?

greenrobot-team (commented):
Found some minor things, LGTM otherwise.

objectbox/lib/src/native/sync.dart
vaind: yeah, noticed the use of the deprecated function so just changed it in the meantime

objectbox/lib/src/relations/to_many.dart
vaind: done

objectbox/lib/src/relations/to_one.dart
vaind: done

objectbox/lib/src/modelinfo/modelproperty.dart
vaind: Nah, AFAIK it just concatenates the string. This is fairly internal error anyway so as long as the model isn't messed up, should never really happen

PR #224, opened 2021-04-23
Simplify data visitors implementation

vaind:
Removes unnecessary code from data visitors implementation - there's no need to do the whole trampoline thing supporting multiple callbacks because data visitors are completely synchronous.


greenrobot-team (commented):
Makes sense and LGTM.

PR #230, opened 2021-04-28
Make sure that runInTransaction() callbacks are always synchronous

vaind:


greenrobot-team (commented):
Maybe document the no async part on the public `Store.runInTransaction` as well?

LGTM.

PR #231, opened 2021-05-03
unsigned integers support

### LINKED ISSUE #85: Unsigned annotation
While there are no standard Dart unsigned integer types, one can still interpret `Int` as unsigned. This is also supported in objectbox-java, using an unsigned annotation. The annotation on a property has an effect when using queries, e.g. ordering the results.

vaind:
closes #85


generator/lib/src/entity_resolver.dart
```diff
@@ -125,6 +125,9 @@ class EntityResolver extends Builder {
       _propertyChecker.runIfMatches(f, (annotation) {
         propUid = annotation.getField('uid')!.toIntValue();
         fieldType = propertyTypeFromAnnotation(annotation.getField('type')!);
+        if (!annotation.getField('signed')!.toBoolValue()!) {
+          flags |= OBXPropertyFlags.UNSIGNED;
+        }
```
greenrobot-team, line 130: Does it make sense to add the flag for any type? E.g. should this error if the property is used on a not supported type?

objectbox/lib/src/native/query/query.dart
```diff
@@ -28,7 +28,8 @@ class Order {
   /// If not specified, the default is case insensitive for ASCII characters.
   static final caseSensitive = 2;
 
-  /// For scalars only: changes the comparison to unsigned (default is signed).
+  /// For integers only: changes the comparison to unsigned. The default is
+  /// signed, unless the property is annotated with [@Property(signed: false)].
   static final unsigned = 4;
```
greenrobot-team, line 33: Should there be a test for this? There doesn't appear to be any.

greenrobot-team (commented):
Simple enough. Found some potential issues.

generator/lib/src/entity_resolver.dart
vaind: Letting core to deal with that at the moment.

vaind: e.g. error-out during model creation

objectbox/lib/src/native/query/query.dart
vaind: should be safe to assume it works, since it's completely handled in the core but I've added a test case regardless.

PR #227, opened 2021-04-26
Improve & clarify size increase FAQ

vaind:

PR #233, opened 2021-05-03
README: add sizes for recommended APK split by ABI.

greenrobot-team:
Addresses issues from #227: use size of APK delivered to users when using recommended App Bundle or APK split by ABI. When using App Bundle the APK generated by the store is about the same size as the split APK built by Flutter.


objectbox/README.md
```diff
@@ -123,16 +123,20 @@ opposite to the recommendation. The setting to disable library compression affec
...
-| .apk uncompressed |         33.81 MiB |      39.60 MiB |     +5.79 MiB |
-| .aab (App bundle) |         15.39 MiB |      17.58 MiB | **+2.19 MiB** |
+A: No, not really. Packaging uncompressed `.so` files is actually better for users. It uses less storage on device as it
+avoids copying .so files from the APK to the filesystem during installation and has the added benefit of making updates
+of your app smaller. For more information about the sizes, see the following table created with 
+ObjectBox v0.14.0 and Flutter v2.0.5 release builds (`flutter build apk` and `flutter build apk --split-per-abi`):
+
+| Release                              | without ObjectBox | with ObjectBox |    Difference |
+| ------------------------------------ | ----------------: | -------------: | ------------: | 
+| Universal APK                        |          15.4 MiB |       39.6 MiB |     +24.2 MiB |
```
greenrobot, line 133: Same as here: https://github.com/objectbox/objectbox-dart/pull/227#pullrequestreview-650105515

objectbox/README.md
```diff
@@ -123,16 +123,20 @@ opposite to the recommendation. The setting to disable library compression affec
...
+| Universal APK (uncompressed)         |          33.8 MiB |       39.6 MiB |      +5.8 MiB |
+| Split APK (uncompressed, arm64-v8a)* |          12.0 MiB |       13.5 MiB |  **+1.5 MiB** |
```
greenrobot, line 135: uncompressed is the default here? Is the command short enough to put into the (*) below?

greenrobot-team: Compressed (so **without** `android:extractNativeLibs="false"`) is the default. Had to add `android:extractNativeLibs="false"` to `AndroidManifest.xml` to get "without ObjectBox" numbers.

greenrobot-team: Note: I was surprised I had to do this [as per documentation](https://developer.android.com/topic/performance/reduce-apk-size#extract-false) uncompressed should be the default based on Android Gradle plugin 3.6.0 or higher being used.

greenrobot-team: Oh, it appears this is due to `minSdkVersion 16` of the Android app template. Apparently there is a bug with loading uncompressed libraries on some Android 6.0 devices: https://flutter.dev/docs/deployment/android#building-the-app-for-release
If raising to `minSdkVersion 23` (Android 7.0), native libs are uncompressed by default.

greenrobot-team: OK, first line is now an APK built with default settings and recommended command (`flutter build apk --split-per-abi`).
However, the `with ObjectBox` column assumes one adds
```
android:extractNativeLibs="true"
tools:replace="android:extractNativeLibs"
```
to turn compression back on (because ObjectBox disables it). I'll probably add a footnote for this, too?

| Release                              | without ObjectBox | with ObjectBox |    Difference |
| ------------------------------------ | ----------------: | -------------: | ------------: | 
| APK (default compressed, arm64-v8a)  |           5.5 MiB |        6.0 MiB |      +0.5 MiB |
| APK (uncompressed, arm64-v8a)*       |          12.0 MiB |       13.5 MiB |  **+1.5 MiB** |
| Fat APK (uncompressed, all ABIs)     |          33.8 MiB |       39.6 MiB |      +5.8 MiB |

greenrobot: Lgtm, foot note sounds good too

greenrobot-team: Changed, will merge.

PR #236, opened 2021-05-07
Align APIs with java/kotlin a little more

vaind:


greenrobot-team (commented):
LGTM, however the current Java query API uses `in/notIn`, the Kotlin extensions `inValues/notIn` (as `in` is a reserved keyword in Kotlin).

The Java/Kotlin preview query API that hasn't shipped since years uses `oneOf/notOneOf`.

So not sure how much this aligns with the Java API as is.

vaind:
h,mm

> LGTM, however the current Java query API uses `in/notIn`, the Kotlin extensions `inValues/notIn` (as `in` is a reserved keyword in Kotlin).
> 
> The Java/Kotlin preview query API that hasn't shipped since years uses `oneOf/notOneOf`.
> 
> So not sure how much this aligns with the Java API as is.

Hmm, was basing the change on the "New Query API" from https://docs.objectbox.io/queries - I guess that's the unshipped one you're referring to. Maybe we should still align so that when it actually does, it will be aligned :) 

`in` is a keyword in dart as well so can't be used, that's why it was `inList` and `inside`.

about `notOneOf` - didn't find that one anywhere in docs (now I see why), but I assumed `noneOf` as the opposite to `oneOf` - maybe we can decide on a single variant for both Dart and Java+Kotlin?

PR #242, opened 2021-05-14
Read speedup

vaind:
The dreaded `Pointer<Uint8>.asTypedList()` being sooo slooow didn't let me rest and it occurred to me to check whether using the same underlying pointer and instead copying the c-memory when reading objects wouldn't end up being faster, And yes, it did, by about a factor of 5 (evens out at about 10 KiB). So here's a PR improving overall read performance (in the flutter bench app) by about 80-100 %. 

As usual, commit-by-commit review recommended

### DIFF
--- a/objectbox/lib/src/native/bindings/flatbuffers.dart
+++ b/objectbox/lib/src/native/bindings/flatbuffers.dart
@@ -96,6 +96,8 @@ class Allocator extends fb.Allocator {
     assert(_data[_index] == data);
     assert(_allocs[_index].address != 0);
 
+    // TODO - there are other options to clear the builder, see how other
+    //        FlatBuffer implementations do it.
     memset(_allocs[_index], 0, data.lengthInBytes);
   }
 
@@ -113,7 +115,7 @@ class ReaderWithCBuffer {
   // See /benchmark/bin/native_pointers.dart for the max buffer size where it
   // still makes sense to use memcpy. On Linux, memcpy starts to be slower at
   // about 10-15 KiB. TODO test on other platforms to find an optimal limit.
-  static const _maxBuffer = 5 * 1024;
+  static const _maxBuffer = 4 * 1024;
   final _bufferPtr = malloc<Uint8>(_maxBuffer);
   late final ByteBuffer _buffer = _bufferPtr.asTypedList(_maxBuffer).buffer;
 
--- a/objectbox/lib/src/native/bindings/flatbuffers.dart
+++ b/objectbox/lib/src/native/bindings/flatbuffers.dart
@@ -1,10 +1,10 @@
 import 'dart:ffi';
-import 'dart:io' show Platform;
 import 'dart:typed_data';
 
 import 'package:ffi/ffi.dart';
 
 import '../../../flatbuffers/flat_buffers.dart' as fb;
+import 'nativemem.dart';
 
 // ignore_for_file: public_member_api_docs
 
@@ -41,12 +41,6 @@ class BuilderWithCBuffer {
 
   Allocator get allocator => _allocator;
 }
-
-// FFI signature
-typedef _dart_memset = void Function(Pointer<Uint8>, int, int);
-typedef _c_memset = Void Function(Pointer<Uint8>, Int32, IntPtr);
-
-_dart_memset? fbMemset;
 
 class Allocator extends fb.Allocator {
   // We may, in practice, have only two active allocations: one used and one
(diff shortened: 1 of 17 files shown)


objectbox/lib/src/native/bindings/flatbuffers.dart
```diff
@@ -98,34 +92,44 @@ class Allocator extends fb.Allocator {
...
   void freeAll() {
     if (_allocs[0].address != 0) calloc.free(_allocs[0]);
     if (_allocs[1].address != 0) calloc.free(_allocs[1]);
   }
 }
+
+/// Implements a native data access wrapper to circumvent Pointer.asTypedList()
+/// slowness. The idea is to reuse the same buffer and rather memcpy the data,
+/// which ends up being faster than calling asTypedList(). Hopefully, we will
+/// be able to remove this if (when) asTypedList() gets optimized in Dart SDK.
```
greenrobot, line 111: Yes, is there some switch to choose the impl (at least on a global level?)

objectbox/lib/src/native/bindings/objectbox.h
```diff
@@ -796,32 +796,32 @@ obx_id obx_cursor_id_for_put(OBX_cursor* cursor, obx_id id_or_zero);
 /// A "put" in ObjectBox follows "insert or update" semantics;
 /// New objects (no pre-existing object for given ID) are inserted while existing objects are replaced/updated.
 /// @param id non-zero
-obx_err obx_cursor_put(OBX_cursor* cursor, obx_id id, const void* data, size_t size);
+obx_err obx_cursor_put(OBX_cursor* cursor, obx_id id, const uint8_t* data, size_t size);
```
greenrobot, line 799: All those switches to uint8_8, that's a dart only change, or is it? What (tool?) requires that and is there an alternative?

greenrobot (commented):
General remark: this is just a first look.

That short before 1.0 release, I'm torn to actually touch this. While results seem to be very promising, touching fundamentals like this is risky (less tested/proven).

greenrobot-team (commented):
LGTM assuming the memcpy code path is as secure as the asTypedList path (no size issues or others I can see, but no idea how asTypedList works).

Not a fan of the manual renames in C binding, but still simple enough.

objectbox/lib/src/native/bindings/flatbuffers.dart
```diff
@@ -98,34 +92,44 @@ class Allocator extends fb.Allocator {
...
+class ReaderWithCBuffer {
+  // See /benchmark/bin/native_pointers.dart for the max buffer size where it
+  // still makes sense to use memcpy. On Linux, memcpy starts to be slower at
+  // about 10-15 KiB. TODO test on other platforms to find an optimal limit.
+  static const _maxBuffer = 5 * 1024;
```
greenrobot, line 116: Is 5 somewhat arbitrary? E.g. why not 8KiB (2*8KB), as a lot of low-level memory stuff works on 4096 boundaries.

objectbox/lib/src/native/bindings/nativemem.dart
```diff
@@ -0,0 +1,25 @@
...
+
+/// memset(ptr, value, num) sets the first num bytes of the block of memory
+/// pointed by ptr to the specified value (interpreted as an uint8).
+final _dart_memset memset =
+    _stdlib.lookupFunction<_c_memset, _dart_memset>('memset');
+
+/// memcpy (destination, source, num) copies the values of num bytes from the
+/// data pointed to by source to the memory block pointed to by destination.
+final _dart_memcpy memcpy =
+    _stdlib.lookupFunction<_c_memcpy, _dart_memcpy>('memcpy');
```
greenrobot, line 14: 1) I did find how this reacts to failures. E.g. does it null or throw?

2) Can we safely assume that 1) it exists and 2) we have access to memcpy?

--> We could have a plan B in place if it fails.

objectbox/lib/src/native/bindings/flatbuffers.dart
greenrobot, line 116: If lookup of memcpy failed (see my comment there), we could set this to zero (and thus disabling memcpy).

objectbox/lib/src/native/bindings/nativemem.dart
```diff
@@ -0,0 +1,25 @@
...
+
+// FFI signature
+typedef _dart_memset = void Function(Pointer<Uint8>, int, int);
+typedef _c_memset = Void Function(Pointer<Uint8>, Int32, IntPtr);
+
+typedef _dart_memcpy = void Function(Pointer<Uint8>, Pointer<Uint8>, int);
+typedef _c_memcpy = Void Function(Pointer<Uint8>, Pointer<Uint8>, IntPtr);
+
+final DynamicLibrary _stdlib = Platform.isWindows
+    ? DynamicLibrary.open('vcruntime140.dll') // required by objectbox.dll
```
greenrobot, line 24: a quick comment why Windows needs special case would be nice

objectbox/lib/src/native/bindings/objectbox.h
greenrobot: Discussed today. Not sure if we'll keep it long term, but yeah, for now...

greenrobot (approved):

objectbox/lib/src/native/bindings/nativemem.dart
vaind: 1. throws
2. we can test on the supported platforms, that's about it
3. plan B would be to use asTypedList(), let's see about the second point first. if it works then I don't see a need to complicate (and slow down) the code with a plan B

objectbox/lib/src/native/bindings/flatbuffers.dart
vaind: Do you have a reference for that 4096 stuff?

8 KiB looks a bit much

objectbox/lib/src/native/bindings/nativemem.dart
greenrobot: Plan B does have to affect performance as outlined on `_maxBuffer` initialization

objectbox/lib/src/native/bindings/flatbuffers.dart
greenrobot: Nothing in particular; 4KB page sizes on Linux, cache boundaries are typically 2^x, ... stuff like this. Thought 8k (or 8000) would still reasonable as its below the 10-15 k threshold. But yeah, not a big difference, nothing we should play around now. It's fine.

objectbox/lib/src/native/bindings/flatbuffers.dart
```diff
@@ -98,34 +92,44 @@ class Allocator extends fb.Allocator {
...
-            .lookupFunction<_c_memset, _dart_memset>('memset');
-      }
-    }
-
     // only used for sanity checks:
     assert(_data[_index] == data);
     assert(_allocs[_index].address != 0);
 
-    fbMemset!(_allocs[_index], 0, data.lengthInBytes);
+    memset(_allocs[_index], 0, data.lengthInBytes);
```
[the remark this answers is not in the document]
vaind: You've created an internal issue for that some time ago. Does that fulfill what you're trying to achieve? If not, maybe move that to GitHub instead? TODOs in the code do tend to get overlooked :/

objectbox/lib/src/native/bindings/flatbuffers.dart
vaind: OK, changing to 4096 for now

objectbox/lib/src/native/bindings/flatbuffers.dart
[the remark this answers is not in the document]
greenrobot: It wouldn't hurt to have that in the code; depending on where you start (issue vs code).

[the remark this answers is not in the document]
vaind: added

objectbox/lib/src/native/bindings/nativemem.dart
vaind: won't have time do do that now. you want to hold it up until then?

objectbox/lib/src/native/bindings/nativemem.dart
vaind: added one

objectbox/lib/src/native/bindings/nativemem.dart
greenrobot: no, let's hope it does not break :crossed_fingers:

PR #243, opened 2021-05-17
Finalizers keep-alive

vaind:
query and property query - add a `reachabilityFence()` after each use of a native pointer with a finalizer attached to prevent premature finalization.

### DIFF
--- a/objectbox/lib/src/native/bindings/bindings.dart
+++ b/objectbox/lib/src/native/bindings/bindings.dart
@@ -122,3 +122,13 @@ late final native_query_close =
     _lib!.lookup<NativeFunction<_native_close>>('obx_query_close');
 late final native_query_prop_close =
     _lib!.lookup<NativeFunction<_native_close>>('obx_query_prop_close');
+
+/// Keeps `this` alive until this call, preventing finalizers to run.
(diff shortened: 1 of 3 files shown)


objectbox/lib/src/native/bindings/bindings.dart
```diff
@@ -122,3 +122,13 @@ late final native_query_close =
...
+
+/// Keeps `this` alive until this call, preventing finalizers to run.
+/// Necessary for objects with a finalizer attached because the optimizer may
+/// mark the object as unused (-> GCed -> finalized) even before it's method
+/// finished executing.
+/// See https://github.com/dart-lang/sdk/issues/35770#issuecomment-840398463
+@pragma('vm:never-inline')
+void reachabilityFence(Object obj) {
+  // Do nothing.
+}
```
greenrobot-team, line 134: The suggestion in the issue was updated to use the `obj`: https://github.com/dart-lang/sdk/issues/35770#issuecomment-842229305

```suggestion
Object reachabilityFence(Object obj) {
  return obj;
}
```

greenrobot-team (commented):
Likely needs updating to match updated recommendation.

Otherwise gnawing my teeth at this, but if finalizers should be in 1.0 I understand this needs to happen.

PR #255, opened 2021-05-26
@Entity(realClass: )  to support some custom code generators

### LINKED ISSUE #229: Support `freezed` entities
Hello,
as far as I know in objectbox-dart we can't create an immutable class.

I have used `@freezed` with "HiveDB"  but I wish we can achieve that with object box

Example:
```dart
part 'priority_entity.freezed.dart';
part 'priority_entity.g.dart';

@freezed
abstract class PriorityEntity with _$PriorityEntity {
  @HiveType(typeId: 2, adapterName: 'PriorityEntityAdapter')
  const factory PriorityEntity({
    @required @HiveField(0) String id,
    @required @HiveField(1) @JsonKey(name: 'org_id') String orgId,
    @required @HiveField(2) String name,
    @required @HiveField(3) bool active,
    @required @HiveField(4) List<PriorityStates> states,
  }) = _PriorityEntity;

  factory PriorityEntity.fromJson(Map<String, dynamic> json) =>
      _$PriorityEntityFromJson(json);
}
```

------------
edit by @vaind: fix the code block

vaind:
closes #229

### DIFF
--- a/generator/integration-tests/part-partof/.gitignore
+++ b/generator/integration-tests/part-partof/.gitignore
@@ -1,0 +1,3 @@
+# start with an empty project, without a objectbox-model.json
+objectbox-model.json
+*.freezed.dart
--- a/generator/integration-tests/part-partof/0.dart
+++ b/generator/integration-tests/part-partof/0.dart
@@ -1,0 +1,12 @@
+import 'dart:io';
+import 'package:test/test.dart';
+import '../test_env.dart';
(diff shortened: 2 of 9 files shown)


greenrobot-team (commented):
LGTM as well as generated code.

It feels like this should have bad side-effects but couldn't think of any (e.g. renaming class should be fine).

PR #256, opened 2021-05-28
Rename query property-meta classes for relations

vaind:
Seems like a good idea to help avoid the confusion in this SO question:

* https://stackoverflow.com/questions/67733318/tomany-relation-gives-type-error-at-link

### DIFF
--- a/generator/lib/src/code_chunks.dart
+++ b/generator/lib/src/code_chunks.dart
@@ -538,9 +538,14 @@ class CodeChunks {
 
       var propCode = '''
         /// see [${entity.name}.${propertyFieldName(prop)}]
-        static final ${propertyFieldName(prop)} = Query${fieldType}Property<${entity.name}''';
-      if (prop.isRelation) propCode += ', ${prop.relationTarget}';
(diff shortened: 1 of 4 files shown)


greenrobot-team (commented):
Is it possible in Dart to introduce a super type both of these are, then let the `link/backlink` method choose the right type (using `is`) on its own? Then drop `linkMany/backlinkMany`?

Anyhow, LGTM.

vaind:
> Is it possible in Dart to introduce a super type both of these are, then let the link/backlink method choose the right type (using is) on its own? Then drop linkMany/backlinkMany?

Yes, thought about that solution as well but that would make it check (and fail) at runtime, while the current solution fails during compilation.

vaind:
> Is that user-facing (not sure what semi-internal here is)? Seems a little long; LinkToOne would be shorter and matching "link".

It's in the generated code. I'm not happy about names either but don't want to change just two of about 10 classes with the same name pattern (they're used for fields in the "underscore" class). Let's not go into the naming discussion in this PR.

PR #264, opened 2021-06-17
Support to-one relation cycles

### LINKED ISSUE #239: Relation cycles
Currently fails with stack overflow: 
```dart
    test('cycles', () {
      final a = RelatedEntityA();
      final b = RelatedEntityB();
      a.relB.target = b;
      b.relA.target = a;
      env.store.box<RelatedEntityA>().put(a);

      final readB = env.store.box<RelatedEntityB>().get(b.id!)!;
      expect(a.relB.targetId, readB.id!);
      expect(readB.relA.target!.id, a.id);
    });
```

We should allow that and also test the same for `ToMany`

vaind:
closes #239

### DIFF
--- a/objectbox/lib/src/native/box.dart
+++ b/objectbox/lib/src/native/box.dart
@@ -166,7 +166,15 @@ class Box<T> {
         throw StateError(
             'Invalid state: can only use _put() on an entity with relations when executing from inside a write transaction.');
       }
-      if (_hasToOneRelations) _putToOneRelFields(object, mode, tx);
+      if (_hasToOneRelations) {
+        // In this case, there may be relation cycles so get the ID first.
+        if ((_entity.getId(object) ?? 0) == 0) {
+          final newId = C.box_id_for_put(_cBox, 0);
+          if (newId == 0) throwLatestNativeError(context: 'id-for-put failed');
+          _entity.setId(object, newId);
+        }
+        _putToOneRelFields(object, mode, tx);
+      }
     }
     _builder.fbb.reset();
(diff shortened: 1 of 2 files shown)


objectbox/lib/src/native/box.dart
```diff
@@ -166,7 +166,17 @@ class Box<T> {
...
       }
-      if (_hasToOneRelations) _putToOneRelFields(object, mode, tx);
+      if (_hasToOneRelations) {
+        // In this case, there may be relation cycles. Therefore, we prepare
+        // the ID beforehand and roll it back on Tx failure.
+        if ((_entity.getId(object) ?? 0) == 0) {
+          final newId = C.box_id_for_put(_cBox, 0);
+          if (newId == 0) throwLatestNativeError(context: 'id-for-put failed');
+          _entity.setId(object, newId);
+          tx.onRollback.add(() => _entity.setId(object, 0));
```
greenrobot, line 176: Not sure if we should start with trying to mimic TX failures in the object model. E.g. other objects, which reference that object, would point to a non-existing object if another TX is retried successfully with the same objects. I don't think we should even try to solve these problems.

vaind: OK, removed

greenrobot-team (commented):
Why is `putMany` not affected?

Otherwise simple fix: assign ID to object before putting its ToOne target to prevent putting by `_putToOneRelFields` should it have a ToOne targeting the original object.

It's a little weird that the assigned ID is overwritten again by the actual put, but OK.

PR #267, opened 2021-06-22
add openStore() to the generated code

### LINKED ISSUE #130: Could not prepare directory: objectbox (30)
I am getting below error when i try to call `Store(getObjectBoxModel());`

Screen Shot 2020-10-14 at 11 37 44 AM

vaind:
Closes #130

### DIFF
--- a/flutter_libs/lib/objectbox_flutter_libs.dart
+++ b/flutter_libs/lib/objectbox_flutter_libs.dart
@@ -1,2 +1,9 @@
-/// This package only contains platform-specific native libraries for flutter.
-/// All the dart code is in package "objectbox".
+/// This package contains platform-specific native libraries for flutter.
+/// See the actual library implementation in package "objectbox".
+
+import 'dart:io';
+
+import 'package:path_provider/path_provider.dart';
+
+Future<Directory> defaultStoreDirectory() async =>
+    await getApplicationDocumentsDirectory();
--- a/flutter_libs/lib/objectbox_flutter_libs.dart
+++ b/flutter_libs/lib/objectbox_flutter_libs.dart
@@ -5,5 +5,6 @@ import 'dart:io';
 
 import 'package:path_provider/path_provider.dart';
 
+/// Returns the default DB dir rooted in the current app's document directory.
 Future<Directory> defaultStoreDirectory() async =>
-    await getApplicationDocumentsDirectory();
+    Directory((await getApplicationDocumentsDirectory()).path + '/objectbox');
--- a/generator/lib/src/code_builder.dart
+++ b/generator/lib/src/code_builder.dart
@@ -8,6 +8,7 @@ import 'package:path/path.dart' as path;
 import 'package:objectbox/internal.dart';
 import 'package:dart_style/dart_style.dart';
 import 'package:source_gen/source_gen.dart';
+import 'package:pubspec_parse/pubspec_parse.dart';
 
 import 'entity_resolver.dart';
 import 'code_chunks.dart';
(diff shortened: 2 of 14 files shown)


objectbox/example/README.md
```diff
@@ -96,7 +90,8 @@ omits the argument to `Store(directory: )`, thus using the default - 'objectbox'
 import 'objectbox.g.dart'; // created by `dart pub run build_runner build`
 
 void main() {
-  final store = Store(getObjectBoxModel()); // Note: getObjectBoxModel() is generated for you in objectbox.g.dart
+  // Store openStore() {...} is defined in the generated objectbox.g.dart
```
greenrobot, line 93: maybe add a quick note on how&when generation is triggered for those who are not familiar with the process?

objectbox/example/README.md
```diff
@@ -96,7 +90,8 @@ omits the argument to `Store(directory: )`, thus using the default - 'objectbox'
...
+  final store = openStore();
```
greenrobot, line 94: Btw.... Thinking of having multiple (separate) models and thus stores at some point in the future, would that clash?
(Might be the same problem with `getObjectBoxModel()` already, but since we touch this...)

vaind: should work - just add an alias to the import

objectbox/example/README.md
greenrobot: Discussed: the error will already occur at the import, which has the info

flutter_libs/lib/objectbox_flutter_libs.dart
```diff
@@ -1,2 +1,9 @@
-/// This package only contains platform-specific native libraries for flutter.
-/// All the dart code is in package "objectbox".
+/// This package contains platform-specific native libraries for flutter.
+/// See the actual library implementation in package "objectbox".
+
+import 'dart:io';
+
+import 'package:path_provider/path_provider.dart';
+
+Future<Directory> defaultStoreDirectory() async =>
```
greenrobot-team, line 8: Some docs on this would be nice, maybe a code snippet on how this should be used?

Edit: this also changes flutter libs to contain Dart code beyond the module loading stuff. Is that an issue? I vaguely remember something about shipping native code only plugins and a Dart issue about this (can't find it right now).

generator/lib/src/code_builder.dart
```diff
@@ -29,6 +30,13 @@ class CodeBuilder extends Builder {
   @override
   FutureOr<void> build(BuildStep buildStep) async {
     // build() will be called only twice, once for the `lib` directory and once for the `test` directory
+    Pubspec? pubspec;
+    try {
+      final pubspecFile = File(path.join(dir(buildStep), '../pubspec.yaml'));
```
greenrobot-team, line 35: Not that familiar with pubspec, but can the flutter libs dependency also come from another file (external variable or import)? Or put differently: can this fail to find flutter libs, even though it's a dependency?

greenrobot-team (commented):
Potential issue. Otherwise LGTM.

flutter_libs/lib/objectbox_flutter_libs.dart
vaind: added docs

> ... beyond the module loading stuff. Is that an issue? 
No issue I can think of. It's absolutely normal to have code in flutter plugins. Ours was actually weird that it didn't have any :) 

Also, I've realized it's missing `/objectbox` at the end, as we suggested in examples previously, so added that as well.

generator/lib/src/code_builder.dart
vaind: I don't know of any other way (which doesn't mean there isn't any, though I doubt it). But it's fine even if it fails - you would just get the plain dart's `openStore()` version and would have to try and figure out why it doesn't open the directory, potentially creating a GH issue here.

flutter_libs/lib/objectbox_flutter_libs.dart
greenrobot-team: > Also, I've realized it's missing /objectbox at the end, as we suggested in examples previously, so added that as well.

Whoops, missed that. Good catch!

PR #268, opened 2021-06-22
Coverage

### LINKED ISSUE #214: Increase test-coverage
We have decent test-coverage with recent code but there is some debt, mostly with early code (queries, property queries). Also, with null-safety enabled, there are now distinct code-branches in the generator for nullable and non-nullable fields so we should test both (current test entities are all nullable). We should:
* [x] introduce test-coverage report to CI - is that possible with dart tooling?
* [x] add nullable entities and cover them with tests
* [x] find blank spaces and cover them (as said above, I suspect that's mostly queries).

vaind:
closes #214


objectbox/test/boxnn_test.dart
```diff
@@ -0,0 +1,88 @@
...
+  setUp(() {
+    env = TestEnv('box-nn');
+    box = env.store.box();
+  });
+
+  tearDown(() => env.close());
+
+  test('CRUD', () {
+    final objects = [
+      TestEntityNN(),
```
greenrobot-team, line 22: Maybe change the `NN` (and `nn`) to at least `NonNull` or similar? Was kind of confused by this as I thought of N as a number placeholder.

objectbox/test/boxnn_test.dart
```diff
@@ -0,0 +1,88 @@
+import 'dart:io';
+import 'dart:typed_data';
+
+import 'package:test/test.dart';
+import 'package:objectbox/objectbox.dart';
+import 'entity.dart';
+import 'entity2.dart';
+import 'test_env.dart';
+
+// Tests non-nullable fields
```
greenrobot-team, line 8: Maybe add a note that this is basically a copy of the regular box tests?

.github/workflows/lint.yml
```diff
@@ -32,4 +32,35 @@ jobs:
...
+          sudo apt-get install lcov
+          make depend
+          make coverage
+      - uses: devmasx/coverage-check-action@v1.2.0
+        with:
+          type: lcov
+          result_path: objectbox/coverage/lcov.info
+          min_coverage: 90
+          token: ${{ github.token }}
+      - uses: actions/setup-dotnet@v1
```
greenrobot-team, line 54: https://github.com/actions/virtual-environments/blob/ubuntu20/20210614.1/images/linux/Ubuntu2004-README.md#net-core-sdk says the latest .NET Core SDK is always installed. No? Edit: k, I guess just copy/pasted from ReportGenerator.

greenrobot-team (commented):
Simple enough, LGTM.

objectbox/test/boxnn_test.dart
vaind: but it is not really, it's much simplified here. added the comment to datetime field test which is pretty much the same

objectbox/test/boxnn_test.dart
vaind: Ok

.github/workflows/lint.yml
vaind: seems to work after removing, thx for the suggestion

PR #269, opened 2021-06-23
reuse TX object (in order to reuse cursors)

vaind:
While looking at test coverage I've noticed each `Transaction` object always held only a single cursor (i.e. the multi-cursor part wasn't ever hit). The investigation made me realize we're always creating a new dart `Transaction` instance (while only a single DB TX is used in the core so that's fine). This PR changes that to reuse the same object internally and reuse cursors.

### DIFF
--- a/objectbox/lib/src/native/store.dart
+++ b/objectbox/lib/src/native/store.dart
@@ -34,6 +34,7 @@ class Store {
   bool _closed = false;
   Stream<List<Type>>? _entityChanges;
   final _reader = ReaderWithCBuffer();
+  Transaction? _tx;
 
   /// absolute path to the database directory
   final String _dbDir;
@@ -263,7 +264,7 @@ class Store {
   }
 
   /// Executes a given function inside a transaction. Returns [fn]'s result.
-  /// Aborts a transaction and rethrows on exception or if [fn] is asynchronous.
+  /// Aborts a transaction and rethrows on exception if [fn] is asynchronous.
   ///
   /// A transaction can group several operations into a single unit of work that
   /// either executes completely or not at all.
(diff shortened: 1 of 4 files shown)


vaind:
> Not sure what exactly this means; e.g. if you use two boxes in a single TX, you would already have two cursors, no?

No. Looking at the code would help here.

> Reuse what object exactly? Dart or native?

Dart. The previous sentence tries to say what is meant by "the object"

objectbox/lib/src/native/store.dart
```diff
@@ -272,8 +273,49 @@ class Store {
...
+      throw UnsupportedError(
+          'Executing an "async" function in a transaction is not allowed.');
+    }
+
+    return _runInTransaction(mode, (tx) => fn());
+  }
+
+  /// Internal only - bypasses the main checks for async functions, you may
+  /// only pass synchronous callbacks!
+  R _runInTransaction<R>(TxMode mode, R Function(Transaction) fn) {
```
greenrobot-team, line 297: What happens if the `mode` changes? Should this function check for that?

greenrobot-team: Hm, maybe not. The database should throw if the mode isn't correct.

greenrobot-team (commented):
LGTM.

Note: also changes transaction to mark successful after executing function (as result needs to be stored in variable anyhow now).

objectbox/lib/src/native/store.dart
vaind: yep, core does

vaind: hmm, actually there was a check for a reused Tx but it got lost in this final version

vaind: added

PR #273, opened 2021-06-24
Query "setParam"

### LINKED ISSUE #209: Add Query setParameter
We should expose setting parameter values, like it is in other languages (for given property or an alias): https://docs.objectbox.io/queries#reusing-queries-and-parameters

vaind:
closes #209


greenrobot-team (commented):
LGTM, maybe needs expanded function docs?

E.g. the alias syntax may imply that it's possible to have multiple properties with the same alias?

vaind:
added docs with examples

PR #276, opened 2021-07-05
flatbuffers - optimization for ASCII strings

vaind:
write directly without UTF16->UTF8 conversion unless needed.

why utf8.encode is so slow? in short, because it: 
1. allocates the whole buffer, up to the maximum size that may be needed if for utf8
2. converts to utf8
3. copies the buffer to a new list with the size of the actually needed buffer
4. the result is then again copied to the output flatbuffer

so two unnecessary allocations and two copies.

This PR takes the opportunistic approach and tries to copy directly to the output buffer, if the given string is ASCII

For ASCII strings, this shows the following improvements in benchmarks
* in plain flatbuffers benchmark - 30-40 % perf increase on Builder)
* in the flutter benchmark app (DB benchmarks - about 20 % increase of `putMany()` operations

For non-ASCII strings, it depends where the first non-ASCII character is, so not sure how to evaluate performance there...

### DIFF
--- a/objectbox/lib/flatbuffers/flat_buffers.dart
+++ b/objectbox/lib/flatbuffers/flat_buffers.dart
@@ -703,18 +703,48 @@ class Builder {
   }
 
   int _writeString(String value) {
-    // TODO(scheglov) optimize for ASCII strings
-    List<int> bytes = utf8.encode(value);
-    int length = bytes.length;
+    // [utf8.encode()] is slow (up to at least Dart SDK 2.13). If the given
+    // string is ASCII we can just write it directly, without any conversion.
+    final originalTail = _tail;
+    if (!_tryWriteASCIIString(value)) {
+      // reset the output buffer position for [_writeUTFString()]
+      _tail = originalTail;
+      _writeUTFString(value);
+    }
+    return _tail;
+  }
+
+  // Try to write the string as ASCII, return false if there's a non-ascii char.
+  @pragma('vm:prefer-inline')
+  bool _tryWriteASCIIString(String value) {
+    _prepare(4, 1, additionalBytes: value.length + 1);
+    final length = value.length;
+    var offset = _buf.lengthInBytes - _tail + 4;
+    for (var i = 0; i < length; i++) {
+      // utf16 code unit, e.g. for '†' it's [0x20 0x20], which is 8224 decimal.
+      // ASCII characters go from 0x00 to 0x7F (which is 0 to 127 decimal).
+      final char = value.codeUnitAt(i);
+      if (char >= 128) {
+        return false;
+      }
+      _buf.setUint8(offset++, char);
+    }
+    _buf.setUint8(offset, 0); // trailing zero
+    _setUint32AtTail(_buf, _tail, value.length);
+    return true;
+  }
+
+  @pragma('vm:prefer-inline')
+  void _writeUTFString(String value) {
+    final bytes = utf8.encode(value) as Uint8List;
+    final length = bytes.length;
     _prepare(4, 1, additionalBytes: length + 1);
-    final int result = _tail;
     _setUint32AtTail(_buf, _tail, length);
-    int offset = _buf.lengthInBytes - _tail + 4;
+    var offset = _buf.lengthInBytes - _tail + 4;
     for (int i = 0; i < length; i++) {
       _buf.setUint8(offset++, bytes[i]);
     }
     _buf.setUint8(offset, 0); // trailing zero
-    return result;
   }
 
   /// Throw an exception if there is not currently a vtable.
--- a/objectbox/test/flatbuffers_test.dart
+++ b/objectbox/test/flatbuffers_test.dart
@@ -83,6 +83,9 @@ void main() {
         as EntityDefinition<TestEntityNonRel>;
 
     final source = TestEntityNonRel.filled();
+    // Test the "dagger" char (0x20 0x20) which may cause problems if
+    // utf16/ascii isn't recognized properly.
+    source.tString = source.tString! + '†asdf';
 
     final fb1 = BuilderWithCBuffer();
     binding.objectToFB(source, fb1.fbb);


objectbox/lib/flatbuffers/flat_buffers.dart
```diff
@@ -703,18 +703,47 @@ class Builder {
...
+    for (var i = 0; i < length; i++) {
+      // utf16 code unit, e.g. for '†' it's [0x20 0x20], which is 8224 decimal.
+      final char = value.codeUnitAt(i);
+      if (char >= 128) {
+        return false;
+      }
+      _buf.setUint8(offset++, char);
+    }
+    _buf.setUint8(offset, 0); // trailing zero
+    _setUint32AtTail(_buf, _tail, value.length);
```
greenrobot-team, line 732: Do I understand correctly that the size is written here last in case above returns `false`?

objectbox/lib/flatbuffers/flat_buffers.dart
```diff
@@ -703,18 +703,47 @@ class Builder {
...
+  // Try to write the string as ASCII, return false if there's a non-ascii char.
+  @pragma('vm:prefer-inline')
+  bool _tryWriteASCIIString(String value) {
+    _prepare(4, 1, additionalBytes: value.length + 1);
+    final length = value.length;
+    var offset = _buf.lengthInBytes - _tail + 4;
+    for (var i = 0; i < length; i++) {
+      // utf16 code unit, e.g. for '†' it's [0x20 0x20], which is 8224 decimal.
+      final char = value.codeUnitAt(i);
+      if (char >= 128) {
```
greenrobot-team, line 726: Maybe add a note that ASCII characters go from 0x00 to 0x7F = 0 to 127 decimal as well?

objectbox/lib/flatbuffers/flat_buffers.dart
```diff
@@ -745,14 +774,15 @@ class Builder {
   /// Additionally allocate the specified `additionalBytes`. Update the current
   /// tail pointer to point at the allocated space.
   @pragma('vm:prefer-inline')
-  void _prepare(int size, int count, {int additionalBytes = 0}) {
+  void _prepare(int size, int count,
+      {int additionalBytes = 0, bool align = true}) {
```
greenrobot-team, line 778: This appears to be never used with `align = false`? Can this change be removed?

objectbox/lib/flatbuffers/flat_buffers.dart
```diff
@@ -703,18 +703,47 @@ class Builder {
...
+    _setUint32AtTail(_buf, _tail, value.length);
+    return true;
+  }
+
+  @pragma('vm:prefer-inline')
+  void _writeUTFString(String value) {
+    final bytes = utf8.encode(value) as Uint8List;
+    final length = bytes.length;
     _prepare(4, 1, additionalBytes: length + 1);
-    final int result = _tail;
```
greenrobot-team, line 710: I wonder why this was cached here, does not seem to change until it is returned?

greenrobot-team (commented):
LGTM based on what was there before and my limited understanding of encoding in FlatBuffers. Just one thing that maybe snuck in there by accident.

objectbox/lib/flatbuffers/flat_buffers.dart
```diff
@@ -703,18 +703,47 @@ class Builder {
...
+  @pragma('vm:prefer-inline')
+  bool _tryWriteASCIIString(String value) {
+    _prepare(4, 1, additionalBytes: value.length + 1);
+    final length = value.length;
+    var offset = _buf.lengthInBytes - _tail + 4;
+    for (var i = 0; i < length; i++) {
+      // utf16 code unit, e.g. for '†' it's [0x20 0x20], which is 8224 decimal.
+      final char = value.codeUnitAt(i);
+      if (char >= 128) {
+        return false;
```
greenrobot, line 727: Could we use the partial result already? E.g. 40 out of 100 char were processed, so we could only have the remaining 60 done with the some existing conversion Dart offers?

objectbox/lib/flatbuffers/flat_buffers.dart
```diff
@@ -703,18 +703,47 @@ class Builder {
   }
 
   int _writeString(String value) {
-    // TODO(scheglov) optimize for ASCII strings
-    List<int> bytes = utf8.encode(value);
-    int length = bytes.length;
+    // [utf8.encode()] is slow (up to at least Dart SDK 2.13). If the given
```
greenrobot, line 706: I'm still wondering if we should have the full utf-16 to utf-8 thing here.  I have the basic conversion in C++ in 30 lines. Needs some additional buffer management.

So, maybe next minor version...

objectbox/lib/flatbuffers/flat_buffers.dart
vaind: yes

objectbox/lib/flatbuffers/flat_buffers.dart
vaind: ah, yeah, leftover from an inbetween solution I tried (converting from the end of the ascii string and switching to utf converter for the rest. Couldn't get it to work because of alignment...

vaind: removed

greenrobot: Yeah, not clear when this would be used. If there's a good case, some docs would be good.

objectbox/lib/flatbuffers/flat_buffers.dart
vaind: Me too but couldn't find any reason so removed.

objectbox/lib/flatbuffers/flat_buffers.dart
vaind: OK, added

objectbox/lib/flatbuffers/flat_buffers.dart
vaind: I've tried that but couldn't get it to work... always seemed to be a couple of bytes off, almost certainly because of alignment (in vectors, etc)

vaind: I'll commit a separate WIP branch for that once this is merged

objectbox/lib/flatbuffers/flat_buffers.dart
vaind: Dart sources are available in the [SDK under BSD license](https://github.com/dart-lang/sdk/blob/master/sdk/lib/convert/utf.dart#L88) and could be adapted. Still not going to be faster than plain ascii writing because it needs the at least one additional allocation for the maximum utf-8 long buffer, right?

greenrobot: > needs the at least one additional allocation for the maximum utf-8 long buffer

Not sure if we could reserve the maximum in FlatBuffers, and later adjust some pointers to "give back" extra unused bytes to the build. That may require transcoding from the end though (because FlatBuffers...)?

An alternative could be to  supply a buffer, which then could be reused to avoid the allocation (requires only the copy then).

objectbox/lib/flatbuffers/flat_buffers.dart
vaind: https://github.com/objectbox/objectbox-dart/tree/draft-fb-string-ascii

greenrobot: PS.: maybe it also could go hand-in-hand with real utf transcoding (my other comment here). E.g. reserve bytes for ASCII in Flatbuffers builder directly, if we need to go UTF, copy partial to a separate buffer and take it from that position...

PR #292, opened 2021-08-03
Relations - json_serializable/freezed support

### LINKED ISSUE #291: JSON to Entity conversion
So probably the closest issue for my problem I found is https://github.com/objectbox/objectbox-dart/issues/251. But the problem is that the explanation is too abstract without examples.

I have asked this question before in Stackoverflow (see link for details) : https://stackoverflow.com/questions/68541015/objectbox-dart-how-to-pass-json-to-entity-class,

The problem is that I need to convert json to multiple related entity classes. The provided solution is to use **final address= ToOne< Address>()**, but the biggest problem I face that I cannot attach Address model to User via constructor, as it expects me to pass ID as int for objectbox.g.dart purposes.

### Describe the solution you'd like

What I would like is that, I could use separate Entity class for business logic purposes and separate Model class for http json conversion which extends entity class. But as address variable is not part of constructor in Entity, I have no clue how to parse it, without having clear examples. 

Issues related suggested to use freezed, but I would like to have as little code generated files as possible. Although I did use @JsonSerializable for json to model conversion. 

Below I provided example what I wanted to achieve, but unfortunately it seems objectbox.g.dart is generating id values based on constructor too, meaning I can't add objects as parameter or ignore it with @Transient(). The other solution I found without any similar examples is provided in (https://stackoverflow.com/questions/68541015/objectbox-dart-how-to-pass-json-to-entity-class), but I'm wondering isn't there any better and cleaner way.
```
@Entity()
class User {
  @JsonKey(ignore: true)
  int id = 0;

  String email;
  String firstName;
  String lastName;
  String? phoneNumber;
  String? birthday;
  @JsonKey(ignore: true)
  final address= ToOne<Address>();

  User({
    required this.email,
    required this.firstName,
    required this.lastName,
    @Transient() Address? address,
    this.phoneNumber,
    this.birthday,
  }) {
    this.address.target = address;
  }
}
```

vaind:
* [x] ToOne support
* [x] ToMany support

closes #291

### DIFF
--- a/generator/lib/src/code_chunks.dart
+++ b/generator/lib/src/code_chunks.dart
@@ -6,8 +6,8 @@ import 'package:pubspec_parse/pubspec_parse.dart';
 import 'package:source_gen/source_gen.dart' show InvalidGenerationSourceError;
 
 class CodeChunks {
-  static String objectboxDart(ModelInfo model, List<String> imports,
-      Pubspec? pubspec) =>
+  static String objectboxDart(
+          ModelInfo model, List<String> imports, Pubspec? pubspec) =>
       """
     // GENERATED CODE - DO NOT MODIFY BY HAND
     
@@ -35,8 +35,7 @@ class CodeChunks {
       ${defineModel(model)}
       
       final bindings = <Type, EntityDefinition>{
-        ${model.entities.mapIndexed((i, entity) => "${entity
-          .name}: ${entityBinding(i, entity)}").join(",\n")}
+        ${model.entities.mapIndexed((i, entity) => "${entity.name}: ${entityBinding(i, entity)}").join(",\n")}
       };
       
       return ModelDefinition(model, bindings);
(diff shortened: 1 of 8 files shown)


objectbox/lib/src/relations/to_many.dart
```diff
@@ -105,10 +119,7 @@ class ToMany<EntityT> extends Object with ListMixin<EntityT> {
 
   @override
   void addAll(Iterable<EntityT> iterable) {
-    iterable.forEach((element) {
-      ArgumentError.checkNotNull(element, 'iterable element');
```
greenrobot-team, line 109: Element null check removed on purpose? Or would the `_count` map throw because it does not support null keys?

If not, element null check is also missing from new constructor.

generator/lib/src/code_chunks.dart
```diff
@@ -374,58 +374,61 @@ class CodeChunks {
...
-          constructorLines.clear();
-          return false;
         } else if (paramType == 'optional') {
           // OK, close the constructor, the rest will be initialized separately.
           return false;
         }
         return true; // continue to the next param
       }
 
+      if (index != null && entity.properties[index].isRelation) {
```
greenrobot-team, line 404: Can this be part of the `index != null` block above that also assigns `paramValueCode`?

greenrobot-team (commented):
LGTM. One potential issue and optimization.

objectbox/lib/src/relations/to_many.dart
vaind: yes, element cannot be nullable (type doesn't allow that)

generator/lib/src/code_chunks.dart
vaind: moved

PR #298, opened 2021-08-17
Forward exceptions from user-provided property converters

### LINKED ISSUE #295: Custom Types property converter swallowing exceptions
Hi there! First of all, thanks for your work on this library!

I had an issue with `query.find` and `query.findFirst` methods that took me a while to figure out and this ticket is intended to making that easier to debug.

If some exception is thrown on the setter of a Custom Types property converter, the exception is swallowed, making it hard to understand what is going on.

For example, say that we have a `User` entity with Custom `Role` type like this:

```dart
import 'package:flutter/material.dart';
import 'package:objectbox/objectbox.dart';

enum Role {
  user,
  admin,
}

@Entity()
class User {
  int id;
  String name;
  Role role;

  User({
    this.id = 0,
    required this.name,
    this.role = Role.user,
  });

  // custom type handling
  int get dbRole {
    return role.index;
  }
  set dbRole(int value) {
    // Exception thrown in the setter:
    throw ErrorDescription('swallowed error??');
  }
}
```

and we add some User records to the db (which will work fine).

Say that we then try to query the users like so:

```dart
// store is opened somewhere else
var box = store.box<User>();
var query = box.query().build();
var count = query.count();
var userIds = query.findIds();
var users = query.find();
var firstUser = query.findFirst();
```

In this case, the `count` and `findIds` queries work as expected, but the `query.find()` call returns an empty list and the `query.findFirst()` call returns null (even when count is > 0). This happens because the exception thrown in the `set dbRole` handler in the user class prevents User objects from being instantiated.

The problem is that this is hard to debug on a real use-case (not explicit exception thrown, but something else going wrong on the setter).

Looks like the exception is being swallowed somewhere in the `runInTransaction` call, but I was unable to find the exact spot.

### Basic info:

- ObjectBox version: 1.1.1
- Flutter/Dart SDK:
    Flutter 2.2.2 • channel stable • https://github.com/flutter/flutter.git
    Framework • revision d79295af24 (9 weeks ago) • 2021-06-11 08:56:01 -0700
    Engine • revision 91c9fc8fe0
    Tools • Dart 2.13.3
- Null-safety enabled: yes
- Reproducibility: always
- OS: macOS Catalina 10.15.7
- Device/Emulator: iOS 14.4 iPhone 12 Pro Max Simulator

### Steps to reproduce

See above, but to sum it up:
1. Have some exception thrown in a custom type converter setter.
2. Query the entity with `find` or `findFirst`

### Expected behavior

The exception that is thrown in the setter should be visible to the developer so that it can be fixed appropriately.

### Actual behavior

The exception is swallowed; `query.find()` returns empty list and `findFirst()` returns `null` even when `query.count()` is > 0.

vaind:
fixes #295

### DIFF
--- a/objectbox/lib/src/native/bindings/data_visitor.dart
+++ b/objectbox/lib/src/native/bindings/data_visitor.dart
@@ -34,9 +34,25 @@ Pointer<NativeFunction<obx_data_visitor>> dataVisitor(
 
 @pragma('vm:prefer-inline')
 Pointer<NativeFunction<obx_data_visitor>> objectCollector<T>(
-        List<T> list, Store store, EntityDefinition<T> entity) =>
+        List<T> list,
+        Store store,
+        EntityDefinition<T> entity,
+        ObjectCollectorError outError) =>
     dataVisitor((Pointer<Uint8> data, int size) {
-      list.add(entity.objectFromFB(
-          store, InternalStoreAccess.reader(store).access(data, size)));
-      return true;
+      try {
+        list.add(entity.objectFromFB(
+            store, InternalStoreAccess.reader(store).access(data, size)));
+        return true;
+      } catch (e) {
+        outError.error = e;
+        return false;
+      }
     });
+
+class ObjectCollectorError {
+  Object? error;
+
+  void throwIfError() {
+    if (error != null) throw error!;
+  }
+}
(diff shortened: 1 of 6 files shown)


greenrobot-team (commented):
LGTM!

Query streams already appear to be fine as exceptions are passed on to the `StreamController`.

PR #316, opened 2021-09-27
Examples: init store in main, access through provider class

greenrobot-team:
- [ ] Update README.
- [ ] Update Sync example.


objectbox/example/flutter/objectbox_demo/lib/objectbox_provider.dart
```diff
@@ -0,0 +1,50 @@
...
+    final store = _store;
+    if (store != null) {
+      return store;
+    } else {
+      throw Exception('Store was not initialized on app launch');
+    }
+  }
+
+  /// Add a note to the Note box.
+  int addNote(Note note) => _box.put(note);
```
greenrobot, line 40: This might result in boilerplate code for each type here. Instead, only expose the store? People get to used to the Box interface without a wrapper around it, I think.

greenrobot-team: This avoids introducing ObjectBox API into the widget code. I would say this is better for testing (e.g. to allow replacing the provider class with a dummy implementation) and is more aligned with the architecture Google is pushing for Android/Flutter apps.

greenrobot-team:
Added an even more basic approach to the example `README` snippet, also explaining why the example adds a wrapper/provider class.

objectbox/example/flutter/objectbox_demo/lib/objectbox_provider.dart
greenrobot: I've seen too many abstractions with little value for the sake of abstraction. Let's discuss in a call tomorrow...

greenrobot-team:
Simplified `ObjectBox` class, updated README and GitBook to match.

objectbox/example/README.md
```diff
@@ -59,27 +59,54 @@ Creating a store
...
+Future<void> main() async {
+  // This is required so ObjectBox can get the application directory
+  // to store the database in.
+  WidgetsFlutterBinding.ensureInitialized();
+
+  objectbox = ObjectBox();
+  await objectbox.init();
+
+  runApp(MyApp());
+}
```
vaind, line 105: What about the following, without the wrapper?
```suggestion
/// Provides access to the ObjectBox Store throughout the app.
late final Store objectbox;

Future<void> main() async {
  objectbox = await openStore();
  runApp(MyApp());
}
```

Also I'd suggest moving `WidgetsFlutterBinding.ensureInitialized()` call to the generated `openStore()`. No need to bother users with that and it should be pretty cheap.

objectbox/example/flutter/objectbox_demo/lib/objectbox.dart
```diff
@@ -0,0 +1,51 @@
...
+import 'objectbox.g.dart'; // created by `flutter pub run build_runner build`
+
+/// Provides access to the ObjectBox Store throughout the app.
+///
+/// Initialized in the apps main function.
+class ObjectBox {
+  Store? _store;
+
+  /// A Box of notes.
+  late final Box<Note> noteBox;
```
vaind, line 11: As an alternative to my suggestion above, this "wrapper" could be part of the generated code but I'd suggest using an async factory instead of constructing an incomplete object and then having to call `init()`.

objectbox/example/flutter/objectbox_demo/lib/objectbox.dart
```diff
@@ -0,0 +1,51 @@
...
+
+  /// Returns the open Store for this app or throws.
+  Store get store {
+    final store = _store;
+    if (store != null) {
+      return store;
+    } else {
+      throw Exception('Store was not initialized on app launch');
+    }
+  }
```
vaind, line 41: e.g. this getter would be unnecessary with an async factory (and a final non-nullable `store`)

vaind (commented):
Some ideas to consider (if you haven't yet)

objectbox/example/README.md
greenrobot-team: Thanks! That was my other suggestion. But decided to use a wrapper class that initializes some re-usable things like boxes and queries. Also to possibly have a single class if a demo test gets added in the future.

As for moving `WidgetsFlutterBinding.ensureInitialized()` into `openStore()`: would this then still work with from a test? But I guess can only call it if `directory` is null.

objectbox/example/flutter/objectbox_demo/lib/objectbox.dart
greenrobot-team: Makes sense as at least right now `init()` does not have any parameters. My initial thought was to prevent any accidental initialization from any other place than the `main()` function, so separated actual init from the getter and made the getter throw.

Do you maybe have a short snippet? I guess this should work using this `??=` syntax? Somewhat like https://github.com/alextekartik/flutter_app_example/blob/master/notepad_sqflite/lib/provider/note_provider.dart#L60-L65

Issue #187, opened 2021-02-03
Failed to load dynamic library (dlopen failed: library "libobjectbox-jni.so" not found)


Katekko:
**Describe the bug**
Just trying to start my store, nothing more

**Basic info:**
 - ObjectBox Dart version: 0.11.0
 - Flutter/Dart SDK: ">=2.7.0 <3.0.0" => Dart SDK version: 2.10.5 (stable) (Unknown timestamp) on "linux_x64"
 - Reproducibility: Create a new project and try to reproduce the very first example on the main page on pubdev
 - Device: Android Emulator - 10.0 (Q) - API 29

Flutter doctor:
```BASH
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 1.22.5, on Linux, locale en_US.UTF-8)
[✓] Android toolchain - develop for Android devices (Android SDK version 30.0.2)
[!] Android Studio (not installed)
[✓] VS Code (version 1.52.1)
[✓] Connected device (1 available)

! Doctor found issues in 1 category.
```

Code that I'm using to try instantiate my store.
The Get.put, is a just my way to put on memory the Store. (See the lib getx)
![image](https://user-images.githubusercontent.com/13632735/106796914-c12cb100-663a-11eb-86da-63a278a89a0e.png)

![Screenshot from 2021-02-03 16-13-55](https://user-images.githubusercontent.com/13632735/106797457-621b6c00-663b-11eb-9cf0-123c6a8bffa6.png)

**Logs, stack traces**
```BASH
StackTrace: 
#0      _open (dart:ffi-patch/ffi_dynamic_library_patch.dart:11:55)
#1      new DynamicLibrary.open (dart:ffi-patch/ffi_dynamic_library_patch.dart:20:12)
#2      loadObjectBoxLib (package:objectbox/src/bindings/bindings.dart:42:26)
#3      C (package:objectbox/src/bindings/bindings.dart:48:41)
#4      new Model (package:objectbox/src/model.dart:18:31)
#5      new Store (package:objectbox/src/store.dart:46:17)
#6      Initializer._initObjectBox (package:arctekko/initializer.dart:70:19)
<asynchronous suspension>
#7      Initializer.init (package:arctekko/initializer.dart:26:13)
#8      _rootRunUnary (dart:async/zone.dart:1198:47)
#9      _CustomZone.runUnary (dart:async/zone.dart:1100:19)
[...]

PR #328, opened 2021-10-27
Add support for Android <= 7.1.1

bm2ilabs:
This pull request for the issue related to #187 
Failed to load dynamic library (dlopen failed: library "libobjectbox-jni.so" not found)


greenrobot-team:
Thanks. Could you provide more details on why you think the added code is necessary to support Android 7.0 and older?
```java
    private static void loadLibrary() {
        if (isLibraryLoaded) {
            return;
        }

        // Loads `libobjectbox-jni.so`.
        System.loadLibrary("objectbox-jni");
        isLibraryLoaded = true;
    }
```

ajit-pandey (approved):

Issue #313, opened 2021-09-22
Flutter application crashes on iOS 15


FredJul:
After upgrading XCode to the very recent version 13.0, our flutter app crashes when doing a box.get(id) with an exception : ArgumentError (Invalid argument(s): Failed to lookup symbol (dlsym(RTLD_DEFAULT, memcpy): symbol not found))

### Basic info (please complete the following information):

- ObjectBox version: 1.2.0
- Flutter/Dart SDK:
     Flutter 2.5.1 • channel stable • https://github.com/flutter/flutter.git
     Framework • revision ffb2ecea52 (5 days ago) • 2021-09-17 15:26:33 -0400
     Engine • revision b3af
- Null-safety enabled: yes
- Reproducibility: always
- OS: macOs BigSur
- Device/Emulator: iOS 15 simulator

Additionally, you can choose to provide more details, e.g. the output of:

* `pub deps --no-dev`
```
Dart SDK 2.14.2
Flutter SDK 2.5.1
sign 0.1.0+1
|-- auto_size_text 3.0.0-nullsafety.0
|   '-- flutter...
|-- collection 1.15.0
|-- cupertino_icons 1.0.3
|-- dartx 0.7.1
|   |-- characters...
|   |-- collection...
|   |-- crypto 3.0.1
|   |   |-- collection...
|   |   '-- typed_data...
|   |-- meta...
|   |-- path...
|   '-- time 2.0.0
|-- dio 4.0.0
|   |-- http_parser 4.0.0
|   |   |-- charcode 1.3.1
|   |   |-- collection...
|   |   |-- source_span 1.8.1
|   |   |   |-- collection...
|   |   |   |-- path...
|   |   |   '-- term_glyph 1.2.0
|   |   |-- string_scanner 1.1.0
|   |   |   |-- charcode...
|   |   |   '-- source_span...
|   |   '-- typed_data...
|   '-- path...
|-- flutter 0.0.0
|   |-- characters 1.1.0
|   |-- collection...
|   |-- meta 1.7.0
|   |-- sky_engine 0.0.99
|   |-- typed_data 1.3.0
|   |   '-- collection...
|   '-- vector_math 2.1.0
|-- flutter_form_builder 6.1.0+1
|   |-- collection...
|   |-- flutter...
|   |-- flutter_localizations...
|   '-- intl...
|-- flutter_localizations 0.0.0
|   |-- characters...
|   |-- clock 1.1.0
|   |-- collection...
[...]

PR #333, opened 2021-11-02
On iOS use Dart API instead of memcpy

greenrobot-team:
Workaround for #313 

If lookup of native memcpy function fails use a Dart implementation instead.
Except in case of `ReaderWithCBuffer` where a somewhat faster fallback implementation already exists (in case its shared buffer is too small it creates a new buffer and returns a view of it).

~To avoid the performance penalty on real devices might detect the iOS simulator. But this seems only possible using a Flutter plugin (based on an Apple flag available at compile time, e.g. see the [Flutter device_info_plus plugin](https://github.com/fluttercommunity/plus_plugins/blob/4e8a3122993c74fe9b57b472ab8c228899ff8e9a/packages/device_info_plus/device_info_plus/ios/Classes/FLTDeviceInfoPlusPlugin.m#L44-L53)).~

### DIFF
--- a/objectbox/lib/src/native/bindings/nativemem.dart
+++ b/objectbox/lib/src/native/bindings/nativemem.dart
@@ -8,10 +8,24 @@ import 'dart:io';
 final _dart_memset memset =
     _stdlib.lookupFunction<_c_memset, _dart_memset>('memset');
 
+/// If the native memcpy function should not be used.
+///
+/// memcpy is not available to Flutter on iOS 15 simulator,
+/// so use Dart API to copy data via asTypedList (which is much slower but works).
+///
+/// https://github.com/objectbox/objectbox-dart/issues/313
+final isMemcpyNotAvailable = Platform.isIOS;
+final _dart_memcpy _memcpyDart = (dest, src, length) {
+  dest
+      .asTypedList(length)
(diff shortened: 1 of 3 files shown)


objectbox/lib/src/native/bindings/nativemem.dart
```diff
@@ -8,10 +8,24 @@ import 'dart:io';
 final _dart_memset memset =
     _stdlib.lookupFunction<_c_memset, _dart_memset>('memset');
 
+/// If the native memcpy function should not be used.
+///
+/// memcpy is not available to Flutter on iOS 15 simulator,
+/// so use Dart API to copy data via asTypedList (which is much slower but works).
+///
+/// https://github.com/objectbox/objectbox-dart/issues/313
+final isMemcpyNotAvailable = Platform.isIOS;
```
greenrobot, line 17: can we instead catch the error of `_stdlib.lookupFunction()` below? Would have a couple of advantages?

greenrobot-team: Yeap, makes more sense, changed!

greenrobot-team: Verified to work on iOS 15 simulator.

greenrobot (approved):

PR #334, opened 2021-11-08
Transaction: replace removed txn_mark_success

greenrobot-team:
Instead of marking write transactions successful or not (`obx_txn_mark_success`) and closing (`obx_txn_close`), use the C functions to mark successful and close (`obx_txn_success`) or just close (`obx_txn_close`).

That just [closing aborts the write transaction is tested](https://github.com/objectbox/objectbox-dart/blob/4a9152238e49182c8db6cc79512fcbfbf8690e6d/objectbox/test/box_test.dart#L517-L530).

### DIFF
--- a/objectbox/lib/src/native/transaction.dart
+++ b/objectbox/lib/src/native/transaction.dart
@@ -13,7 +13,7 @@ import 'bindings/helpers.dart';
 
 /// Represents a native transaction - it is bound to a current thread so never
 /// use with asychcronous code, or more specifically, never `await` before
-/// calling [close].
+/// calling [successAndClose] or [abortAndClose].
 @internal
 class Transaction {
   final Store _store;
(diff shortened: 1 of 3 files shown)


objectbox/lib/src/relations/to_many.dart
```diff
@@ -210,9 +210,8 @@ class ToMany<EntityT> extends Object with ListMixin<EntityT> {
             throw UnimplementedError();
         }
       });
-      if (ownedTx) tx.markSuccessful();
     } finally {
-      if (ownedTx) tx.close();
+      if (ownedTx) tx.commitAndClose();
```
greenrobot, line 214: Looks like changed semantics to me; before it aborted unless it reached tx.markSuccessful().

objectbox/lib/src/native/transaction.dart
```diff
@@ -33,37 +33,18 @@ class Transaction {
...
-      }
-    } else {
-      close();
-    }
-  }
-
+  /// Commits a write transaction by marking it successful.
+  /// Read transactions are just closed.
   @pragma('vm:prefer-inline')
   void commitAndClose() => _finish(true);
```
greenrobot, line 39: Btw, this is private API, right? Otherwise, "commit" would be misleading and inconsistent to other other languages.

greenrobot-team: The class is marked `@internal`, but language wise it's not private. Anyhow, what would be a better name? Still makes sense to use it.

objectbox/lib/src/relations/to_many.dart
greenrobot-team: Totally right, guarded code even does explicitly throw. Added catch clause.

objectbox/lib/src/native/transaction.dart
greenrobot: `success()` as in C++?

Issue #297, opened 2021-08-14
@Unique: replace on conflict flag


dimitribocquet:
I am new to Objectbox, but it seems to fit to my needs, except for one thing: **upsert data**.
I fetch some items to my API server (300 items) and they have a UUID, with **@Unique**. Sometimes, these items are updated and I would like to fetch/update them, by their UUID.

FYI this is what I'm doing currently, I don't know if it's the best way:

```dart
List<CardModel> cards;

cards.forEach((card) {
        card.extension.target = extension;

        CardModel? cardInDb = _cardRepository.findByUuid(card.uuid);

        if (cardInDb != null) {
          card.id = cardInDb.id;
        }
});

_cardRepository.box.putMany(cards);
```

After some researches, I've seen that Objectbox Java has a OnConflict flag for @Unique:  

`@Unique(onConflict = ConflictStrategy.REPLACE)`
 See https://github.com/objectbox/objectbox-java/issues/509

It would be great if we have the same thing in Dart/Flutter.

greenrobot-team:
This can now be tested with the latest preview release: `1.3.0-dev.0`

See https://docs.objectbox.io/entity-annotations#unique-constraints for an example.

greenrobot-team:
This is now available as part of the `1.3.0` release.

PR #338, opened 2021-11-09
Support Unique conflict resolution strategy replace.

greenrobot-team:
#297

Add support for the replace conflict resolution strategy 
- to be used on at most one property,
- if Sync is enabled all `@Unique` properties must use `REPLACE` strategy (and with above rule at most one can exist).

# Tasks
- [x] Check if sync-enabled entity, all unique properties use `REPLACE`.
- [x] Add tests.
  - `build_runner` tests seem not possible (e.g. to verify errors on misconfigurations), so tested manually.

### DIFF
--- a/generator/lib/src/entity_resolver.dart
+++ b/generator/lib/src/entity_resolver.dart
@@ -257,6 +257,11 @@ class EntityResolver extends Builder {
       }
     }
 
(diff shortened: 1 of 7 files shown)


greenrobot (approved):
lgtm

PR #341, opened 2021-11-28
Add an option to configure generator's output_dir

### LINKED ISSUE #312: Allow declaring directory for generated files
`objectbox-model.json` and `objectbox.g.dart` are generated in _lib_ directory and there's no way of changing it which is inconvenient if you follow some pattern when structuring your project files. Allow ObjectBox users to adjust it in `build.yaml` configuration e.g.:
```
targets:
  $default:
    builders:
      objectbox:
        options:
          output_dir: lib/generated/or/wherever/you/want/it/to/be
```

vaind:
closes #312 

- [x] Check failed test: `custom` directory needs to exist.
- [x] Imports in generated code are wrong (assumes generated code file is in top-most directory).
- [x] Add documentation.
- [x] Clean up integration test (currently just copy/paste from `final` test).

### DIFF
--- a/generator/integration-tests/config/.gitignore
+++ b/generator/integration-tests/config/.gitignore
@@ -1,0 +1,4 @@
+# start with an empty project, without a objectbox-model.json
+objectbox-model.json
+objectbox.*
+testdata
--- a/generator/integration-tests/config/1.dart
+++ b/generator/integration-tests/config/1.dart
@@ -1,0 +1,48 @@
+import 'dart:io';
+import 'dart:typed_data';
+
+import 'package:test/test.dart';
+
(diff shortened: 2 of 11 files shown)


greenrobot-team (commented):
Thanks!
Changed to prefix imports with one or more `../` to fix imports in generated code file.
Fixed imports in test, reduced to minimal required.
Added documentation.
@vaind Maybe have a look at the import changes, can you think of a better solution?

vaind:
> Maybe have a look at the import changes, can you think of a better solution?

LGTM. I've made a minor change - not sure it's needed but doesn't hurt anything I guess

PR #361, opened 2021-12-31
Format the sources and run a check in CI

vaind:
... to avoid unrelated changesets in PRs


vaind:
merging in order to use the formatted sources in other PRs

greenrobot-team:
Shouldn't the package analysis (`pana`) job catch this? I guess it does not look at the example code?

PR #364, opened 2022-01-06
fixed link to the example

awaik:


greenrobot-team:
The current relative link works just fine, no?

awaik:
Hi, I noticed that on the pub.dev
Screenshot 2022-01-11 at 07 36 11
Screenshot 2022-01-11 at 07 36 27

greenrobot-team:
Thanks! Actually the issue is that the `objectbox` subfolder is missing from the URL created by pub.dev, then the link would work. Using an absolute link would only work for the example link, but the README link at the top of https://pub.dev/packages/objectbox/example is also broken the same way.

I wonder how other packages are doing this. Does it depend on which working directory is used to publish maybe?

awaik:
You are right, didn't notice that generated link doesnt work.
_Does it depend on which working directory is used to publish maybe?_
Maybe it is. Usually this link just autogenerated.
In you case on the github link is
https://github.com/objectbox/objectbox-dart/blob/main/objectbox/example/flutter/objectbox_demo
and on the pub.dev
https://github.com/objectbox/objectbox-dart/blob/master/example/flutter/objectbox_demo

greenrobot-team:
At least fixed the example link by adapting the release script to add an absolute one for pub.dev: 5226820a6ecc56665ac45ad3f4f54e1541d7da28

Still need to figure out how to fix the auto-generated README link (probably by publishing from a different working directory?).

greenrobot-team:
Link to example README will be fixed with next release as well: 60fc8ee1

PR #366, opened 2022-01-08
add MIT LICENSE

SajadRahimi1:


greenrobot-team:
The packages are licensed under Apache 2.0, see https://github.com/objectbox/objectbox-dart/blob/main/objectbox/LICENSE

PR #375, opened 2022-01-31
Fix Flutter CI, migrate to lints from pedantic

greenrobot-team:
- [ ] pedantic is discontinued and new checks use lints. Need to fix analysis issues to get perfect package score again.
- [x] Enable cache for https://github.com/subosito/flutter-action
  - Cache key does not include version, so not sure if this will always get latest Flutter from channel. Need to check with next Flutter release.


greenrobot-team:
Flutter changes merged to `main`.

For lint updates see #382.

PR #382, opened 2022-02-15
Migrate from pedantic to lints

greenrobot-team:
pedantic is discontinued and replaced by lints. Update to it and update code to pass new rules.

- [x] Fix analysis issues to get perfect package score again.


greenrobot-team:
Actually the issue is that `pana` now ignores the excluded files in `analysis_options.yaml`. https://github.com/dart-lang/pana/issues/1013

Instead add ignore rules.

Issue #307, opened 2021-08-29
Freezed support with non-assignable id


ilikerobots:
In #229, the example supporting freezed entities uses assignable ids.  Since freezed demands entities to be immutable, attempts to use non-assignable ids fails in https://github.com/objectbox/objectbox-dart/blob/e587529f3a32ce7769e9a53a24ea4afafdd6d970/objectbox/lib/src/native/box.dart#L222 with errro
```
Unhandled Exception: Invalid argument(s): Field Model.id is read-only (final or getter-only) and it was declared to be self-assigned. However, the currently inserted object (.id=0) doesn't match the inserted ID (ID 1). You must assign an ID before calling [box.put()].
```

Is it possible to use non-assignable ids with immutable entities currently?  If not, would it be beneficial to make it so, e.g. by modifying the code above to use a copyWith or some other mechanism to set newly assigned ids?

vaind:
Not sure how we could do this. The contract of `box.put()` (and `putMany()`) is that it sets the ID on the given object, if it was previously zero. This is especially important with relations (and relation cycles) so that when the related data is being put, the existing objects aren't reinserted always as new ones (that would result in an infinite recursion).

ilikerobots:
Ok, thanks.  If I understand correctly,then, a project that uses objectbox and immutable models is required to fully manage IDs independently, right?

lampian:
Using Equatable requires entity class to be immutable.
This then require the Id field to be final, which currently is not supported.
Please consider providing a way to update the entity as a whole with all fields as final.
The solution suggested by @tomwyr is pointing in the right direction - not sure what side effects there would be.

vaind:
@tomwyr - unfortunately that still won't satisfy the current "contract" of `box.put()` (or `putMany()`) which says it updates the given object. Besides this being possibly relied upon in client apps, it's also used when working with relations, requiring the relation target object instance to be updated with the inserted ID. Getting around this may require some more thought... 

Anyone interested in this change, please upvote the original issue so the interest can be tracked when prioritizing development.

rafuck:
https://github.com/objectbox/objectbox-dart/pull/393

raj457036:
This works btw! ( note: this will make your properties mutable else everything will be same )

```dart
@Freezed(
      addImplicitFinal: false,
)
sealed class YourClass with _$YourClass {
    @Entity(realClass: YourClass)
    factory YourClass({
        @Default(0) @Id() int localId,
[...]

PR #393, opened 2022-03-04
Immutable entities with copyWithId setter

rafuck:
Immutable entities support (see [Issue-307](https://github.com/objectbox/objectbox-dart/issues/307))
```dart
@Entity()
class TestEntityImmutable {
  @Id(useCopyWith: true)
  final int? id;

  final int payload;

  TestEntityImmutable copyWith({int? id, int? payload}) =>
      TestEntityImmutable(
        id: id,
        payload: payload ?? this.payload,
      );

  TestEntityImmutable({this.id, required this.payload});

  TestEntityImmutable copyWithId(int newId) =>
      (id != newId) ? copyWith(id: newId) : this;
}
```

### DIFF
--- a/generator/lib/src/code_builder.dart
+++ b/generator/lib/src/code_builder.dart
@@ -209,6 +209,7 @@ class CodeBuilder extends Builder {
     propInModel.name = prop.name;
     propInModel.type = prop.type;
     propInModel.flags = prop.flags;
+    propInModel.generatorFlags = prop.generatorFlags;
     propInModel.dartFieldType = prop.dartFieldType;
     propInModel.relationTarget = prop.relationTarget;
 
--- a/generator/lib/src/code_chunks.dart
+++ b/generator/lib/src/code_chunks.dart
@@ -125,7 +125,8 @@ class CodeChunks {
       id: ${createIdUid(property.id)},
       name: '${property.name}',
       type: ${property.type},
-      flags: ${property.flags}
+      flags: ${property.flags},
+      generatorFlags: ${property.generatorFlags}
       $additionalArgs
     )
     ''';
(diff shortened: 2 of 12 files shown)


greenrobot-team:
Thanks, we'll have a look!

greenrobot-team:
From a quick look this breaks the contract that put modifies the same object [as mentioned in the original issue](https://github.com/objectbox/objectbox-dart/issues/307#issuecomment-1002467103).

rafuck:
> From a quick look this breaks the contract that put modifies the same object [as mentioned in the original issue](https://github.com/objectbox/objectbox-dart/issues/307#issuecomment-1002467103).
Yes, put and putMany (and etc) change ID for objects before send it to storage. In the case of immutability (in this PR) it is imposible to change immutable objects, and it sends to storage new objects, constructed by copyWith. But it is obvious to the client code that an immutable entity should not change. And if such code uses immutability, it should not rely on side effects. In other cases, everything works as before.

greenrobot-team:
Again thanks for this! I guess you already figured out that relations are going to be a problem as well.

We'll have a look at this in more detail and if it makes sense to integrate.

rafuck:
> Again thanks for this! I guess you already figured out that relations are going to be a problem as well.
Yes. My bad: I don't think about relations at the start.. It is really lot of work to make relations in immutable case!
> We'll have a look at this in more detail and if it makes sense to integrate.
Thank you

rafuck:
Ping?

greenrobot-team:
@rafuck Haven't looked at this, yet.

erdzan12:
up

gombal88:
any update in the case of immutable classes and autoincrementing id?

PR #410, opened 2022-04-16
image in objectbox

chiggosVikash:
how we can set and get image from object box dart


greenrobot-team:
I guess you already asked this on Stack Overflow. This (GitHub pull requests) is not the right place for this.

Issue #424, opened 2022-05-18
Freezed entities don't work well with objectbox_generator


atarana:
ObjectBox Generator doesn't generate `objectbox-model.json` and `objectbox.g.dart` with freezed entities.

As a workaround I found that using `flutter pub run build_runner watch --delete-conflicting-outputs` and changing just one of the entity's `@Id(assignable: )` from `true` to `false` and vice-versa does the trick sometimes. 

### Basic info (please complete the following information):

- ObjectBox version: 1.5.0
- Flutter/Dart SDK: Flutter 3.0.0 Stable - Dart 2.17.0
- Null-safety enabled: yes
- Reproducibility: always
- OS: macOs BigSur 
- Device/Emulator: Galaxy A50

Additionally, you can choose to provide more details, e.g. the output of:

* `pub deps --no-dev`:
```
Dart SDK 2.17.0
Flutter SDK 3.0.0
testProj 0.8.2+43
|-- animated_check 1.0.5
|   '-- flutter...
|-- animated_text_kit 4.2.1
|   |-- characters 1.2.0
|   '-- flutter...
|-- another_flushbar 1.10.28
|   '-- flutter...
|-- auto_route 4.0.1
|   |-- collection 1.16.0
|   |-- flutter...
|   |-- meta 1.7.0
|   |-- path 1.8.1
|   '-- universal_html 2.0.8
|       |-- async...
|       |-- charcode...
|       |-- collection...
|       |-- csslib 0.17.1
|       |   '-- source_span...
|       |-- html 0.15.0
|       |   |-- csslib...
|       |   '-- source_span...
|       |-- meta...
|       |-- source_span 1.8.2
|       |   |-- collection...
|       |   |-- path...
|       |   '-- term_glyph 1.2.0
|       |-- typed_data...
|       '-- universal_io 2.0.4
|           |-- collection...
|           |-- crypto...
|           |-- meta...
|           '-- typed_data...
|-- cached_network_image 3.2.0
|   |-- cached_network_image_platform_interface 1.0.0
|   |   |-- flutter...
|   |   '-- flutter_cache_manager...
|   |-- cached_network_image_web 1.0.1
|   |   |-- cached_network_image_platform_interface...
|   |   |-- flutter...
|   |   '-- flutter_cache_manager...
|   |-- flutter...
|   |-- flutter_cache_manager 3.3.0
|   |   |-- clock...
|   |   |-- collection...
|   |   |-- file 6.1.2
[...]

PR #472, opened 2022-09-27
Freezed: be sure it is run before objectbox generator

vlado-rudenok:
### Description
The PR  fixes an issue when ObjectBox Generator is run before freezed, what makes impossible to generate database scheme.

Related issues
https://github.com/objectbox/objectbox-dart/issues/424
https://github.com/Milad-Akarie/auto_route_library/issues/1086

The issue is reproducible provided the project contains freezed, [objectbox](https://github.com/objectbox/objectbox-dart) and [auto_route](https://github.com/Milad-Akarie/auto_route_library) dependencies.


greenrobot-team:
Thanks! However, I don't think it's a good idea to have workarounds explicitly for other libraries. So not sure about this (also, would this even work if no freezed files are present?).

PR #530, opened 2023-04-20
Support long (8 byte integer) type

### LINKED ISSUE #529: Treat int as 8 bytes
### Describe the solution you'd like

Current implementation of `PropertyType.int` seems to be [32 bit integer](https://github.com/objectbox/objectbox-dart/blob/main/objectbox/lib/src/annotations.dart#L87). However, dart integers are usually 64 bit integers although it depends on the platform. Thus, it can happen that objectbox truncates the integer to 32 bits only and lose some information.

### Describe alternatives you've considered

I considered using `byteVector` which I believe would work, but the data that I'm dealing with is 64 bit integers, so I would much prefer if objectbox can handle 64 bit integers in the first place.

Thanks!

semin-park:
closes #529

There already is `OBXPropertyType.Long`, so this PR adds this to the `PropertyType`.

### DIFF
--- a/objectbox/lib/src/annotations.dart
+++ b/objectbox/lib/src/annotations.dart
@@ -86,6 +86,9 @@ enum PropertyType {
   /// size: 4-bytes/32-bits
   int,
 
+  /// size: 8-bytes/64-bits
+  long,
+
   // dart type=int, size: 8-bytes/64-bits
   // no need to specify explicitly, just use [int]
   // long,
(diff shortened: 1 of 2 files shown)


MatheusPatricioo:
,

greenrobot-team:
Thanks! Not sure, why this was not closed. But see https://github.com/objectbox/objectbox-dart/issues/529#issuecomment-1531070782

PR #562, opened 2023-09-21
static methods can be called from an instance

leslmosnk:
static methods can be called from an instance of class with ObjectBox queries.

### DIFF
--- a/generator/lib/src/code_chunks.dart
+++ b/generator/lib/src/code_chunks.dart
@@ -724,17 +724,29 @@ class CodeChunks {
               'Unsupported property type (${prop.type}): ${entity.name}.$name');
       }
 
-      var propCode = '''
+      var staticPropCode = '''
         /// see [${entity.name}.${propertyFieldName(prop)}]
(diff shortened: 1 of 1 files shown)


greenrobot-team:
Thanks! However, I don't really see why this is needed. Can you explain how this is useful to you?

leslmosnk:
> Thanks! However, I don't really see why this is needed. Can you explain how this is useful to you?

This way I managed to extend `Object` with `Object_`. So I can easily call a query parameter from an instance of `Object`. 

(I later found out `export 'objectbox.g.dart' show Object_` is enough for my purposes)

greenrobot-team:
But why can't the `Object_` fields be used in your case? We chose explicitly not to add them to the model class to keep it clean (e.g. to avoid any conflicts with existing code).

leslmosnk:
Its very possible to use `Object_`, but I feel like it only used whenever I also use my `Object`. So i'd rather have a static in `Object` that would hold my queries instead of having to import and call `Object_` every time.

PR #575, opened 2023-10-26
Namespace

ologunB:
add condition for agp compatibility


greenrobot-team:
Thanks, however the plugin package is built with Android Plugin 7.1. So what exactly is the reason for this change?

https://github.com/objectbox/objectbox-dart/blob/733aae4acaa98b9e6fb2ad864da7aadce6244388/flutter_libs/android/build.gradle#L12

Anyhow, we should update the plugin packages to use a newer Android Plugin.

greenrobot-team:
Thanks again, this was resolved as part of changes to our Flutter plugins. See #581.

Issue #522, opened 2023-03-25
Cmake warning issue when building windows desktop app


KhaelM:
Running or building Windows desktop app after objectbox dependencies importation shows following warnings: 

```
CMake Warning (dev) at C:/Program Files/Microsoft Visual Studio/2022/Community/Common7/IDE/CommonExtensions/Microsoft/CMake/CMake/share/cmake-3.24/Modules/FetchContent.cmake:1264 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
Call Stack (most recent call first):
  flutter/ephemeral/.plugin_symlinks/objectbox_flutter_libs/windows/CMakeLists.txt:58 (FetchContent_Declare)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

### Basic info (please complete the following information):

- ObjectBox version: objectbox: ^2.0.0 | objectbox_flutter_libs: any
- Flutter/Dart SDK: 3.7.8
- Null-safety enabled: yes
- Reproducibility: always
- OS: Windows 11
- Device/Emulator: Windows x64

greenrobot-team:
Thanks for reporting! As noted this is just a warning about a deprecated CMake feature and should not prevent the build from succeeding. Anyhow, we should look into resolving this.

greenrobot-team:
It looks like this [is known for Flutter](https://github.com/flutter/flutter/issues/116866) and they [fixed this in their tests and examples](https://github.com/flutter/flutter/pull/125502/files).

I was about to do the same, but with Flutter 3.16.2 I can not longer reproduce the error message (e.g. `flutter clean` and then `flutter build windows`).

greenrobot-team:
Never mind, [can still see it on CI](https://github.com/objectbox/objectbox-dart/actions/runs/7103033563/job/19334760261) (probably Cmake stuff is cached somewhere `flutter clean` does not clean).

greenrobot-team:
[...]

PR #579, opened 2023-11-09
Fix Windows Build CMP0135 Policy Deprecation

ikelly963:
Addresses #522 

Regarding this Policy: https://cmake.org/cmake/help/latest/policy/CMP0135.html
Setting Policy: https://cmake.org/cmake/help/latest/command/cmake_policy.html#setting-policies-explicitly

- [x] I've read and followed [CONTRIBUTING.md](https://github.com/objectbox/objectbox-dart/blob/main/CONTRIBUTING.md)


ikelly963:
First time contributing to a public GitHub repo. Please let me know if I need to do something additional.

greenrobot-team:
Thank you! I'll have a look when I have time.

For me: from a first look the used option is not added until 3.24, but the minimum supported CMake is lower. Check how that works on older versions. Also the docs note that [enabling specific policies is discouraged](https://cmake.org/cmake/help/latest/command/cmake_policy.html#setting-policies-explicitly:~:text=While%20setting%20policies%20individually%20is%20supported%2C%20we%20encourage%20projects%20to%20set%20policies%20based%20on%20CMake%20versions).

ikelly963:
Ok, great. I'll take a look at those docs and figure out why these tests are failing later, potentially today.

greenrobot-team:
@ikelly963 Don't worry about that. `pana` fails because I need to still adjust the threshold and `coverage` just has no permission to upload the report because this is a branch outside of this repo.

greenrobot-team:
The Flutter folks have a better fix. But I'm not sure it's still needed. See https://github.com/objectbox/objectbox-dart/issues/522#issuecomment-1849712951

PR #695, opened 2024-11-26
Fix examples for JDK 21 (Android Studio Ladybug)

greenrobot-team:
Public PR to run GitHub Actions.


hagerf:
Would this PR make the warnings below that objectbox currently give me disappear?
`warning: [options] source value 8 is obsolete and will be removed in a future release
warning: [options] target value 8 is obsolete and will be removed in a future release
warning: [options] To suppress warnings about obsolete options, use -Xlint:-options.`

greenrobot-team:
@hagerf No, [it won't](https://github.com/objectbox/objectbox-dart/actions/runs/12134912957/job/33832960168?pr=695). Thanks for reporting!

This warning occurs due to using JDK 21 to compile.

But we can make changes to the Flutter plugins to suppress it, made an internal note about that (`objectbox-dart#127`)!

If it bothers you, you can try to suppress the warning yourself. You likely need to add something like below to the relevant Gradle script in the `android` folder (note the below is for a KTS build script, it likely needs to be adapted to work in a Groovy build script typically used by Flutter projects):

```
// TODO Convert to Groovy to use in build.gradle (this is for a build.gradle.kts file)
tasks.withType(JavaCompile::class.java).configureEach {
    // Suppress JDK 21 warning about deprecated, but not yet removed, source and target value 8 support
    options.compilerArgs.add("-Xlint:-options")
}
```

Long term we need to use the Java 11 level to compile (Flutter [appears to also prepare for that](https://github.com/flutter/flutter/commit/b55290697705851d35d4d31258881be09da3ca36)), but that has other complications (like minimum supported Android SDK level).

hagerf:
Thanks for the quick and detailed answer! 
I assumed that it wouldn't. And thanks for the suggestions. Even though the warnings are a bit bothersome, I'm not a fan of suppressing warnings as they actually give useful information, in this case it means that some dependencies should be upgraded to compile with a newer version of Java. And even though I understand that raising the Android SDK level might not be something you want to do lightly, that's the real solution to this. Is there some thread or task to follow on Github for that?
Thanks again

greenrobot-team:
As part of this change I plan to suppress the warnings. And no, there is currently only an internal task to update to Java 11 level once it makes sense.

hagerf:
Is this planned to be released soon? And maybe we could add [this](https://github.com/objectbox/objectbox-dart/pull/712) to this as well? I'm experiencing issues with updating some dependencies because object box is using old source_gen versions.

greenrobot-team:
@hagerf Please avoid asking for updates, we won't provide any until things are ready. It just takes time away from actually working on these tasks.

Issue #705, opened 2025-01-21
Support source_gen 2 and analyzer 7


Shajk0:
Hey guys

I wanted to test objectbox, but I get a dependency error.
Seems like a too old version of source_gen is used and so I have a conflict with other packges.
Could you please update this to source_gen to version 2.0.0

`Because no versions of objectbox_generator match >4.0.3 <5.0.0 and objectbox_generator 4.0.3 depends on source_gen ^1.0.0, objectbox_generator ^4.0.3 requires source_gen ^1.0.0.`

Edit:
Found another conflict:
`Because no versions of objectbox_generator match >4.0.3 <5.0.0 and objectbox_generator 4.0.3 depends on analyzer >=5.2.0 <7.0.0, objectbox_generator ^4.0.3 requires analyzer >=5.2.0 <7.0.0.
And because riverpod_lint >=2.6.4 <3.0.0-dev.0 depends on analyzer ^7.0.0, objectbox_generator ^4.0.3 is incompatible with riverpod_lint >=2.6.4 <3.0.0-dev.0.`

greenrobot-team:
Thanks for reporting! We will test with the newer versions, make required changes and then allow them in pubspec.

As for the first message the relevant part is missing ("is incompatible with...") so I can't advice. But the solution is the same as for the second conflict:

Change your pubspec to allow versions of the conflicting dependecy, like for `riverpod_lint` allow lower than version 2.6.4. Then Flutter (or rather pub) will find versions that work together for you.

simeonApproppo:
I have the same issue when I try to used the actual version of freezed: (2.5.8)
```
And because objectbox_generator >=0.13.0+1 depends on source_gen ^1.0.0 and freezed >=2.5.8 <3.0.0-0.0.dev depends on source_gen ^2.0.0,
  freezed >=2.5.8 <3.0.0-0.0.dev is incompatible with objectbox_generator.
```

greenrobot-team:
@simeonApproppo If you are saying my recommendation to allow more versions does not work, I updated [my comment above](https://github.com/objectbox/objectbox-dart/issues/705#issuecomment-2606371103) to include a workaround for that case.

simeonApproppo:
Downgrading to freezed 2.5.7 is working.. But I hope that you will release a version, that depends on a actual version of source_gen soon. :-)

mike-500:
[...]

PR #712, opened 2025-03-03
chore: update dependencies

IFLyer:
Update analyzer to ">=6.0.0 <8.0.0" - Update dart_style to ^3.0.1 and use latest language version - Update source_gen to ^2.0.0


luigi198:
@greenrobot-team could we get a check here? I think this PR fix the source_gen issue we are experiencing [here](https://github.com/objectbox/objectbox-dart/issues/705)

PR #798, opened 2026-04-02
Change Admin bindUri 127.0.0.1 to 0.0.0.0

azhon:
Change 127.0.0.1 to 0.0.0.0 will allow access from all devices on the local network by default.

### DIFF
--- a/objectbox/lib/src/native/admin.dart
+++ b/objectbox/lib/src/native/admin.dart
@@ -38,7 +38,7 @@ class Admin implements Finalizable {
   static bool isAvailable() => C.has_feature(OBXFeature.Admin);
 
   /// Creates an ObjectBox Admin associated with the given store and options.
-  Admin(Store store, {String bindUri = 'http://127.0.0.1:8090'}) {
+  Admin(Store store, {String bindUri = 'http://0.0.0.0:8090'}) {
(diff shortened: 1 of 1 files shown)


azhon:
This is possible, but due to the small screen size of mobile devices, ObjectBox Admin Web is not ideal for viewing data on mobile devices. Therefore, it's more commonly viewed on a computer.
Furthermore, using `debugImplementation` does not pose any security issues in the release mode.

greenrobot-team:
To browse Admin on your development machine, forward the port using adb: https://docs.objectbox.io/data-browser#browse-data-on-your-dev-machine

azhon:
This is also a solution

greenrobot-team:
> but configure the URL to bind to

The URL that Admin binds to is already configurable? (Edit: which [azhon already blogged about](https://azhon.blog.csdn.net/article/details/159771330?spm=1001.2014.3001.5502).)

```dart
final admin = Admin(env.store, bindUri: 'http://0.0.0.0:8090');
```

greenrobot-team:
After internal discussion closing this as the URL is already configurable and defaulting to bind to all interfaces is just too dangerous (such as if accidentally shipping an Admin-enabled build to production).

PR #805, opened 2026-05-10
build(android): migrate Kotlin compiler options

Tinnci:
## Draft status

This PR keeps the existing Kotlin Gradle Plugin application for current repository CI compatibility, while migrating deprecated Android `kotlinOptions` blocks to top-level `kotlin { compilerOptions { ... } }`.

This is intentionally not the full AGP 9 built-in Kotlin removal. Removing `kotlin-android` should be handled separately only when this repository is ready to require AGP 9 built-in Kotlin.

## Scope

Only Android Gradle build configuration is changed. No Dart APIs, Android manifests, SDK versions, or runtime code are changed.

## Verification

- `rg "kotlinOptions"` on touched Android Gradle files returns no matches.
- `rg "compilerOptions"` confirms JVM target configuration exists.
- `git diff --check`

### DIFF
--- a/flutter_libs/android/build.gradle
+++ b/flutter_libs/android/build.gradle
@@ -22,7 +22,6 @@ allprojects {
 }
 
 apply plugin: "com.android.library"
-apply plugin: "kotlin-android"
 
 android {
     namespace = "io.objectbox.objectbox_flutter_libs"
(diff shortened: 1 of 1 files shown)


greenrobot-team:
Thanks! We're planning to modernize the examples to use the latest Flutter templates anyhow, so this will be resolved as a part of that. There is no need to make this change right now?

Tinnci:
Sounds good, since you are planning to modernize the templates anyway, I will close this PR. Thanks!

PR #807, opened 2026-05-13
Widen the analyzer package upper version to <14

seanolsen:
Widen the analyzer package upper version to <14


greenrobot-team:
Thanks! However, we'll address this as part of our usual release cycle. Does the current version restriction prevent you from updating any other package dependencies of your project?

For our reference, this is internal issue `objectbox-dart#184`

seanolsen:
yes. there is a bunch of packages that I can't update because of analyzer version limitation in objectbox

PR #822, opened 2026-07-06
Web platform support: Store, Box and queries on IndexedDB

### LINKED ISSUE #185: Web-capable Box API; based on IndexedDB

mechaadi:
Closes #185 (phases 1–3 of the approach discussed there).

This PR makes objectbox, objectbox_generator-generated code and objectbox_flutter_libs compile and run on the web platform, under both dart2js and dart2wasm. It builds on the conditional-import scaffolding that was added for this purpose back in #189 (the empty lib/src/web/ placeholders and per-file facades), and follows the layering sketched by @greenrobot in [this comment](https://github.com/objectbox/objectbox-dart/issues/185#issuecomment-913246667), with pure Dart taking the place of the WASM core: since the native core can't currently target WASM (mmap), the web implementation is a self-contained Dart engine behind the existing public API.

### What works on web
**Store/Box:** all CRUD operations with PutMode semantics, monotonic ID assignment (ids of removed objects are not reused, also across sessions), @Unique enforcement incl. ConflictStrategy.replace (UniqueViolationException), ToOne/ToMany/backlinks via the unchanged shared relation code, runInTransaction with rollback on error (undo log), watch<T>()/entityChanges streams, Store.attach, in-memory databases via memory: paths.

**Queries**: the full condition set (string ops with case-sensitivity flags, integer/date/double/bool/byte-vector/string-vector conditions, isNull/oneOf/between, and/or groups), order() flags, offset/limit, find*/count/remove/stream, query parameters with alias support, property queries (min/max/sum/average/distinct/find), relation links (link/backlink/linkMany/backlinkMany, nested, relationCount), and nearestNeighborsF32 evaluated as an exact brute-force scan with findWithScores (Euclidean/cosine/dot-product; Geo throws).

**Persistence**: an in-memory working set (preserving the synchronous API) flushed to IndexedDB in batched transactions via a write-behind queue; data, unique indexes, relations and ID sequences survive reload. navigator.storage.persist() is requested best-effort.

### Design notes
**New public API:** Store.ready, a future that completes when persisted data is loaded (completes immediately on native). Generated Flutter openStore() awaits it.

**FlatBuffers on web:** dart2js has no ByteData.get/setInt64, so serialization cannot use package:flat_buffers as-is. The web build uses a vendored copy (Apache-2.0 headers retained) with JS-safe 64-bit accessors, exposed through a conditional export package:objectbox/flatbuffers.dart; native platforms re-export the real package through the same entrypoint, so types are identical and native behavior is unchanged. Generated code imports the shim.

**Generator:** model IDs/UIDs and retired-UID lists are emitted as runtime-parsed strings (IdUid.fromString('1:…')) — UIDs are 64-bit values that dart2js rejects as literals above 2^53. IdUid's former (1 << 63) - 1 range check wrapped negative under JS 32-bit shift semantics and is now a lower-bound check (the upper bound is enforced by int.parse on the VM).

**Shared code hygiene:** modelinfo/relations no longer transitively import dart:ffi (EntityDefinition.objectFromData moved to a native-only extension; the two HNSW enums are mirrored in enums.dart like the other OBX enums already were; PutMode moved into the facade alongside TxMode).

`Minimum SDK is now 3.4 (dart:js_interop / package:web). The ffigen bindings are pinned // @dart=2.19 until regenerated (Dart 3 class-modifier rules).
`
### Not supported on web
Sync and Admin (isAvailable() returns false so guard code keeps working), Store.fromReference, isolate APIs (runAsync/runInTransactionAsync run on the same thread), Geo vector distance. Queries scan the in-memory records (no index acceleration). Durability is write-behind: writes flush within a microtask, so a hard browser crash can lose the final moments; close() flushes. Multi-tab access is not coordinated.

### Testing

1. All 213 tests in objectbox_test pass on the VM — native behavior is unchanged.
2. 30 browser tests (engine + queries + relations + persistence-across-reopen) pass in Chrome under both dart2js and dart2wasm; happy to contribute them as a package/CI job in this PR or a follow-up.

Validated end-to-end in a production Flutter app (entities with unique indexes, relations and schema-evolution history): `flutter build web --release` and Android builds green, app verified working in the browser on IndexedDB.

I'm happy to split this into smaller PRs (the commits are already staged that way: compile-support → engine → queries → generator fixes), adjust naming/placement, or rework the FlatBuffers approach if you'd prefer a different mechanism.


greenrobot-team:
Thank you! This is a very large change made without discussing with us first. So I'm not sure if or when this will get merged.

I assume this was mostly created using a coding tool?

mechaadi:
@greenrobot-team Thanks for taking a look, and fair point i almost forgot to read that part in contribution guidelines, my bad, I should have commented on #185 before starting, apologies for that. 

I'm running this as a fork in production either way, so please treat the PR as a starting point for discussion rather than something waiting on a clock.

Yes, the implementation was largely written with an AI coding agent (Claude Code), working under my direction. I'd rather be upfront about that so you can calibrate review accordingly. What I'd offer as counterweight is the verification, which was the focus throughout:

1. All 213 tests in `objectbox_test` pass unchanged, native behavior was the hard constraint at every step.
2. The branch adds a `web_test package` with 30 browser tests (CRUD of all property types, PutMode semantics, unique enforcement, ToOne/ToMany/backlinks, transaction rollback, persistence across reopen, the full query surface, vector search with scores), run in CI under both dart2js and dart2wasm.
3. `generator/test.sh` passes, including the model-evolution case.
4. It's validated end-to-end in a production Flutter app (entities with unique indexes, relations, and schema-evolution history) on web and Android.
 
Additionally, the design follows the layering you sketched in #185 back in 2021, with pure Dart standing in for the wasm core, and reuses the conditional-import scaffolding from #189. There's a design write-up covering the architecture, semantics (including where web is deliberately weaker, e.g. write-behind durability), and the dart2js pitfalls encountered, I am more than happy to add it to the PR if its useful in any way.

mechaadi:
@greenrobot Indeed it was fable 5.

Glad this can be used as a foundation, i am also more than happy to contribute in that or anything else if you would need my assistance.

Great library!

ftr-tuta:
I’m curious about this direction and whether there’s anything useful I could help with on the Dart side.

After reading the discussion on #822, I was wondering if a smaller foundational PR could be useful for the future official Web/WASM support, without implementing a separate Dart/IndexedDB engine.

The idea could be to focus only on preparing the Dart package for an official Web backend, for example:

* make shared runtime/model code independent from `dart:ffi`;
* use the existing `native/` vs `web/` platform seams and modernize them for `dart:js_interop`;
* move platform-neutral API types such as Sync credentials/states out of the native FFI implementation;
* make sure model IDs/UIDs can cross the Web boundary without losing 64-bit precision;
* leave `web/store.dart`, `web/box.dart`, `web/query.dart` and `web/sync.dart` as adapters for the planned official TypeScript/WASM ObjectBox implementation, rather than reimplementing the database in Dart.

The goal wouldn’t be to provide Web support itself, but to make `objectbox-dart` easier to connect to the real ObjectBox Web runtime later, including Sync.

Would this be an interesting direction for a small base PR, or is there another area around the upcoming Web support where contributions would be more useful?

